PR #8, opened 2015-07-31
Remove `try` in float parser.

albertnetymk:
According to this post
(http://blog.ezyang.com/2014/05/parsec-try-a-or-b-considered-harmful/),
the error message could be confusing if using `try`. `"3.a"` is one case
to expose this behavior.


mrkkrp:
Thank you, I'll need to watch out for this in future. I'm planning concept of QuickCheck test suite now. I think we could check quality of error messages as part of it too.

----

I'm granting you write access to the repository, this may be of little use for major undertakings, since it's better to discuss such changes via issues and PRs, but minor corrections will be done more easily if you have write access.

PR #30, opened 2015-09-10
Add a few synthetic benchmarks

### LINKED ISSUE #7: Why don't we benchmark the stuff?
It might be a good idea to benchmark at least basic things. Not sure I can do all the stuff myself, but let it hang here, maybe someone could help in future and write some benchmarks. I think `criterion` library can be used.

tulcod:
This code benchmarks the "string" primitive and various non-primitive combinators.
The code coverage of these benchmarks is not 100%: new benchmarks should be added
as relevant performance questions are discovered.

This closes #7.

### DIFF
--- a/benchmarks/Main.hs
+++ b/benchmarks/Main.hs
@@ -27,12 +27,168 @@
 -- any way out of the use of this software, even if advised of the
 -- possibility of such damage.
 
+{-# LANGUAGE CPP #-}
+
 module Main (main) where
 
 import Criterion.Main
+
+import Text.Megaparsec
+
+-- Configuration parameters.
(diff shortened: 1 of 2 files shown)


mrkkrp:
Thank you for this, this is highly appreciated. I'll try to find time and adapt this for Parsec to compare performance of Megaparsec and Parsec.

tulcod:
The [picoparsec](https://bitbucket.org/blamario/picoparsec) project also has a benchmark suite, which includes various real-world benchmarks, and compares parsec+picoparsec+attoparsec. Though I couldn't get it to build.

Issue #27, opened 2015-09-07
Using parsers with monad transformers


neongreen:
*(See the previous discussion [here](https://github.com/mrkkrp/megaparsec/issues/18#issuecomment-138137081).)*

`ParsecT` lets us combine parsing with things such as keeping state, accessing configuration, etc. However, any monad you embed into `ParsecT` won't be affected by Parsec's backtracking properties:

~~~ haskell
{-# LANGUAGE FlexibleContexts #-}

import Control.Monad.State
import Text.Parsec hiding (State)

trying s = modify (++ ["trying " ++ s])

test :: ParsecT String () (State [String]) Char
test = (trying "'a'" >> char 'a')
   <|> (trying "'b'" >> char 'b')

main = print $ flip execState [] $ runParserT test () "" "b"
-- prints:
--   trying 'a'
--   trying 'b'
~~~

In this example having non-backtracking state was desirable. Sometimes having backtracking state is desirable – for instance, my usecase is parsing a list of rules, where each rule is merely a named list of patterns, and patterns can reference earlier-defined patterns (which is why all parsed patterns are added to state); if a single pattern in a rule can't be parsed I want to throw the rule away completely and forget that I ever parsed it, which means forgetting about all successfully-parsed patterns in the rule. (Sorry if the example is unclear or seems contrived.)

One way to have backtracking state is to use Parsec's built-in “user state”. Unfortunately, this solution isn't extensible, as Parsec provides no built-in writer, reader, or any other useful monad you could think of.

Another way is to use

~~~ haskell
StateT MyStateType Parser a
~~~

which, if you aren't familiar with `StateT`, is another way of saying

~~~ haskell
MyStateType -> Parser (a, MyStateType)
~~~

which hopefully makes it clearer why this variant would preserve backtracking; if it doesn't, look at the definition of `<|>` for `StateT`:

~~~ haskell
m <|> n = StateT $ \ s -> runStateT m s <|> runStateT n s
~~~

Here both actions are applied to the same state, so it's impossible that both parsers' modifications would end up in the final state.

At the moment this solution isn't viable either, because all Parsec's parsers work in `ParsecT` and can't be used in `StateT`. So, we have to use `lift`:

~~~ haskell
{-# LANGUAGE FlexibleContexts #-}

import Control.Applicative
import Control.Monad.State
import Text.Parsec hiding (State, (<|>))
import Text.Parsec.String

trying s = modify (++ ["trying " ++ s])

[...]

PR #33, opened 2015-09-21
Add some tests for monad transformers (see #27)

neongreen:
This brings coverage to 82%; it could be made even higher by replicating more tests for `ReaderT` and `IdentityT`, but it feels like... cheating? Also I can't think of any non-trivial tests for `getParserState` and `updateParserState` that would be related to `StateT`.


mrkkrp:
@neongreen, well, it's not cheating :-) When you have trivial things that are copied of already checked things, you can add trivial tests for them. We cannot foresee future changes, and extra testing cannot hurt.

PR #37, opened 2015-09-22
Improve the docs a bit

neongreen:


neongreen:
I have compared Coveralls logs from this build and the previous one, and apparently the culprit is this line: [before](https://coveralls.io/files/1229717080#L532), [after](https://coveralls.io/files/1232087978#L537). I have no idea why it changed. Perhaps it has to do something with Quickcheck and different seeds used in each case?

mrkkrp:
Yeah, fluctuations happen with QuickCheck, it's OK.

----

In Megaparsec, certain conventions in documentation are used. You should explain why do you think you should *locally break* them and why your changes are *necessary*.

1. Why do you think `@...@` is better than currently used `>`. If you want to change it, you need to change it everywhere, otherwise you're introducing inconsistency and entropy into docs without good reason.

2. Sorting of combinations by utility is questionable. They are already sorted alphabetically (with exception to deprecated combinators). Again you're increasing entropy, because these combinators are sorted the same way in `Text.Megaparsec` export. Now you have one order in one file and another order elsewhere. Sorting is already pretty useless for these things because there is no real criterion which can be used to sort them. The only necessary thing is to preserve uniform order everywhere and it's broken now. Also note that parsers themselves are ordered in the source file the same way as in import section of the file. Again, you change one thing and don't see the other. Please use something like interactive rebase to eliminate this reordering completely.

3. The same thing with `|p|` notation vs `@p@`. Any real benefit? Again, you're introducing entropy because if you have really good reason (not whim) to change the markup, change it *everywhere*.

4. Please don't create special section for deprecated combinators. The fact that I mark this section in comments doesn't mean that there should be such section in documentation. These are normal combinators, there is no reason why they should have separate section. For example, you could be looking for combinators in documentation, but would you look for “deprecated combinators” on purpose? Probably not.

Once we have consensus on these points I will continue reviewing, right now changes seem to be mostly unnecessary. This doesn't mean that you don't have good things here, some changes are good, but they should be separated from unnecessary ones.

mrkkrp:
Don't bother to correct anything, since commits are actually well separated I'll just cherry pick the ones I like.

mrkkrp:
I've cherry picked the first and the last of the commits. Feel free to argue if you think other should be included as well, but please address my criticism from the first comment first.

neongreen:
> Why do you think `@...@` is better than currently used `>`. If you want to change it, you need to change it everywhere, otherwise you're introducing inconsistency and entropy into docs without good reason.

I only used it when I wanted to add hyperlinking. `>` doesn't allow hyperlinking.

mrkkrp:
@neongreen, Yes, but is that necessary? All relevant functions are mentioned and hyperlinked in the text already.

mrkkrp:
@neongreen, can you try to fix these things I mentioned via interactive rebase, remove already applied and not-desirable (like creating separate section for deprecated parsers) commits? There are still good corrections we could have in Megaparsec.

neongreen:
> Yes, but is that necessary? All relevant functions are mentioned and hyperlinked in the text already.

I have noticed that when I'm reading documentation, my attention level is low when I'm reading text and high when I'm reading a code sample (this is by no means universal behavior, but I believe it might be common enough). So, if there's some function I haven't seen before, I'll only notice it (and reach for a hyperlink) when I get to the code sample, not during skimming the text.

Generally, I don't see any reason why something *shouldn't* be hyperlinked when it can (well, apart from the fact that currently hyperlinking also means highlighting, and highlighting only relevant parts is more useful than highlighting everything). Duplication is an excellent thing.

neongreen:
> The same thing with `|p|` notation vs `@p@`. Any real benefit? Again, you're introducing entropy because if you have really good reason (not whim) to change the markup, change it everywhere.

Well, in this case I was actually reducing entropy. All other places were using `@p@`.

(Not to mention that `|p|` notation isn't even supported by Haddock anymore.)

mrkkrp:
@neongreen, Yes, that is a good change. Sorry, I initially misread it as if you are using `|p|`, not eliminating it. Don't know why this is in Megaparsec source, should be `@p@`, of course.

neongreen:
> They are already sorted alphabetically (with exception to deprecated combinators).

Alphabetical sorting is useless. Even the “but finding things in a sorted list in easier” argument doesn't apply – the easiest way to find something on a web page is to use Ctrl-F, and the easiest way to find something in a source file is to use either that or “jump to definition”.

> Sorting is already pretty useless for these things because there is no real criterion which can be used to sort them.

When I'm reading something, I want to know how relevant/useful it is. If it's not, I'm going to skim it automatically; if it is, I'm going to pay attention because I might need it later and so I'd better remember it. “First item on the list is more important than the last” is an implicit way of signalling usefulness.

`++` is the 1st function of `Data.List` and `genericReplicate` is the last one. The top bar of <apple.com> has “Mac” and “iPad” sorted before “TV” and “Music”, despite the whole list being only 7 entries. On <microsoft.com> we have a 7-entry list beginning with “Windows” and ending with “Bing”. Those lists are extremely short, there's no conceivable reason why they *should* be sorted by “how often will it be clicked”, and yet they are. Given that those people actually perform usability testing, I would conclude that it's reasonable to make sorting by priority the default order (not the Only Ultimate Order, merely the default one!) unless there's some reason to do otherwise.

What benefits does alphabetical order have, apart from “now anyone who wants to add a combinator to `Text.Megaparsec.Combinators` has only one place where it could be inserted” (which doesn't seem useful)?

neongreen:
> These are normal combinators, there is no reason why they should have separate section.

It's is a good way to convey “if you're reading to get an overview of the library, you can stop here and switch to some other module”.

mrkkrp:
@neongreen, is has no particular benefits, although if you know what you're looking for you can estimate where in file it will approximately be. But how do you know that `option` is more important than `count`? I mean, when you talk about these short lists where important things go first, I agree that some thought was put in ordering of the items. The items can be ordered because the companies has their own views on what is more important. They have *point of view*. This doesn't necessarily coincide with point of view of end user, actually.

In other words, this is rather matter of taste. The principle itself is good (important things first), but it's difficult to apply here. I cannot say which of the combinators is more used, I don't have statistic. We can come up with countless variants of ordering and I don't think it changes anything.

----

The combinators are marked as deprecated in very visible way and there is not so many of them.

neongreen:
> But how do you know that `option` is more important than `count`?

https://github.com/alogic0/parsec-doc/blob/275d24c3f5d32732a91ca083be7f7105bc706b27/hackage-analysis/parsec3-import-stats.txt

neongreen:
Well, or even https://github.com/alogic0/parsec-doc/blob/275d24c3f5d32732a91ca083be7f7105bc706b27/hackage-analysis/combined-import-stats.txt. I have to say that it didn't occur to me to search for it when I was actually making the pull request, so I resorted to guessing instead. Approximate order is still better than alphabetical order.

mrkkrp:
@neongreen, This may be of interest. I'll return to this once tab width is flexible.

neongreen:
> The combinators are marked as deprecated in very visible way and there is not so many of them.

Okay. I definitely don't think it's important enough to be worth arguing over.

mrkkrp:
Ordering is not particularly important either. After all it simply doesn't matter much especially in this sort of document. I doubt it will make any difference at all. Those who want to read docs of a function will be able to do it just as well. Alphabetic sorting gives at least some hints, I don't like when things are displayed in random order (what you propose is essentially random order). This would be better if we had several groups of combinators (like in `Data.List`), then we could sort the groups. But there are not so many combinators, and those that we have are almost equal in importance (or what they do in principle, they combine). It's not `Data.List`, there you have sections like “searching list”, etc.

neongreen:
> remove already applied and not-desirable (like creating separate section for deprecated parsers) commits?

Done.

neongreen:
(Or should I have also rebased on top of current master?)

mrkkrp:
@neongreen, Well if you had rebased on top of current master, you would have some problems perhaps. If you want this sort of rebase (which would be better of course), you should open another pull request. But don't do this. I'm in process of changing source in way that may conflict with your edits. I think I'll commit flexible tab width and then commit some changes you propose.

mrkkrp:
@neongreen, 90% of your current propositions get into master. I've edited a couple of lines myself and added short doc-string for `showToken` function. For some reason I decided to avoid links to Haskell report. If you press hard-enough they may get through though :-D

neongreen:
> For some reason I decided to avoid links to Haskell report. If you press hard-enough they may get through though :-D

Nah, I'll probably need all my power of persuasion to get `parse'` renamed to `parseMaybe` and get a `parse'` that throws an exception, so I'll avoid arguing for the Haskell report links.

mrkkrp:
@neongreen, for this open a separate issue. At any rate I'm done for today.

mrkkrp:
@neongreen, renaming `parse'` to `parseMaybe` is a good thing since there are already several flavors of `parse`, so longer name is justified (it's more descriptive). About variation that throws an exception: you should tell what sort of exception and why it should be in the library.

Issue #42, opened 2015-09-30
Add base-4.7.0.0 compatibility


bkaestner:
This should be rather easy. It's more or less the question whether you actually want backwards-compatibility, and whether you want to use `{-# LANGUAGE CPP #-}` for the `MIN_VER` macro or "redundant import" warnings.

I just tried a quick 'n dirty patch and imported either `Control.Applicative` (17 files\*) or `Data.Foldable (Foldable, foldMap)` (1 file), which passed all tests. Note that the import of `Control.Applicative` isn't necessary in all of those files, e.g. all `parseFromFile` could be written with infix `fmap` instead of `<$>`.

If you're interested in a patch, I can probably prepare a proper one later today.

\* Bug9.hs is the only file that needs `hiding (Const)` due to `data Expr = Const …`.

mrkkrp:
Answering your question, yes, I'm interested in that patch, it would be great. It's now apparent that some users would benefit from compatibility with `base-4.7.0.0` and the library would be more widely used.

Enable `CPP` extension globally in Megaparsec, in `megaparsec.cabal`, then please use `MIN_VER` to add the “missing” imports conditionally. When you have pull request ready we'll discuss other details if necessary.

mrkkrp:
OK, Travis [says](https://travis-ci.org/mrkkrp/megaparsec) that it builds with GHC 7.8 now. Although I Cabal 1.18 required different flag (`--enable-library-coverage` instead of `--enable-coverage`), so I had to put if/then/else statement in `.travis.yml` file. I had to put it all in one line, what a disaster. Closing this.

PR #45, opened 2015-09-30
Add compatibility to GHC 7.8.x

bkaestner:
This PR basically introduces a bunch of CPP `#if`s to import additional operators from `Control.Applicative` (in most cases) or `Data.Foldable` (in one case). See #42 for the original issue.

I didn't change the existing `import Control.Applicative` lines, since they would lead to redundant import warnings on GHC 7.10.x, and instead put the `#if MIN_VERSION` makros at the end of all imports in all files. in all files that used only `<$>` in one place, I replaced `<$>` by infix `fmap` to keep the number of changes per file a little bit smaller.

While I was at it, I also removed the "recent version of `base`" comment in the readme, since it works now with an older one (by the way, base-4.6 compatibility seems possible, currently experimenting).

Things left to do:

- [ ] change Travis configuration to automatically check compatibility with 4.7.x

-----

## Remarks

- the fixity declaration of `/=\` in tests/Utils.hs lead to some issues. I've added a comment, which will prevent backslash-before-newline trouble.
- while the current version works, it introduces many changes to many files. The next version of base shouldn't break anything (at least regarding the `Control.Applicative` part), but one should probably think of a less error prone strategy, like a custom prelude (so that there's only a single point of CPP).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,14 +35,11 @@ Megaparsec is different from Parsec in the following ways:
 * Megaparsec's code is clearer and doesn't contain “magic” found in original
   Parsec.
 
-* Megaparsec looks into the future, it does not contain code that serves for
-  compatibility purposes, it also requires more recent version of `base`.
-
 ## Megaparsec vs Parsec
 
 There are good reasons to use Parsec:
 
-* You need to work with legacy code or with older versions of GHC (< 7.10).
+* You need to work with legacy code or with older versions of GHC (< 7.8).
 
 And that's it. In other cases you should prefer Megaparsec for your own
 sake. If you think you have a reason to use Parsec other than listed here,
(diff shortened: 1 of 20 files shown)


mrkkrp:
Great work, thank you. I'm a bit sleepy right now, so I will return to this tomorrow to check and merge. While you're at it, please amend that last commit and also edit “(< 7.10)” → “(< 7.8)” (that's in the section “Megaparsec vs Parsec”) changing description of the commit accordingly.

Yep, we'll need to make Travis test GHC 7.8 too, but it's trivial. I'm not sure whether we should use custom Prelude and real benefits of such solution. I don't find conditional imports of `Control.Applicative` particularly ugly (in any sense) right now.

bkaestner:
> I don't find conditional imports of Control.Applicative particularly ugly (in any sense) right now.

Yeah, the conditional imports aren't that bad. However, the conditional imports/redefines on base-4.6.x.x. Well, see #46 :D.

> While you're at it, please amend that last commit 

Err, I think I've misread your comment as `git rebase master --interactive` and squashed both commits into one. Not sure whether you wanted that or only the `7.10 -> 7.8` amended.

mrkkrp:
Yes, I meant only that last commit dealing with `README.md` file. But in its current state it's also fine. I'm not sure compatibility with `base-4.6.x.x` is a priority, however. We will talk about this tomorrow in dedicated thread.

Issue #34, opened 2015-09-21
An embedded tutorial


neongreen:
Megaparsec might be able to replace Parsec much faster if it has a tutorial (not just a list of distinctions between `parsec` and `megaparsec`) – this way instead of everyone saying “okay there's Megaparsec but maybe use Parsec instead because all tutorials cover it and they aren't that different” we might end up with “okay there's Parsec but Megaparsec comes with an extensive tutorial so look at it instead”. So, I offer to write such a tutorial, and then we could put it in `Text.Megaparsec.Tutorial`.

(The reason I'm making this an issue instead of just silently writing it and then making a pull request is that I don't want a repetition of the situation with [Tekmo and lens tutorial](https://github.com/ekmett/lens/pull/552). So, if e.g. @mrkkrp has strong opinions on how it should be written, it would be better to hear them early on instead of ending up with something written in [my usual idiosyncratic style](http://artyom.me/aeson) and then having the pull request rejected.)

mrkkrp:
@neongreen, Great idea. I was thinking about creating a web page (we can use GitHub project pages to host it) describing Megaparsec. This would contain manual, instructions how to switch for Parsec users, etc. I plan to address the need for this “auxiliary” material after initial release of the library.

As I understand you're proposing something built-in into source. This is interesting, although I have to admit (to my shame) that I see such thing for the first time. I don't have anything against it and thus such a pull request will be probably merged, not rejected. You can start working on the tutorial, but I encourage you open pull request earlier so I and other people can comment as it progresses and result won't be something surprising.

neongreen:
> built-in into source

Yep, like [the pipes tutorial](https://hackage.haskell.org/package/pipes-4.1.6/docs/Pipes-Tutorial.html) that everyone seems to love.

mrkkrp:
@neongreen, Do you have any plan of things/sections to include in this tutorial?

neongreen:
* a tiny introduction to monadic parsers
* writing some parser (don't know what yet)
* adding `<?>` stuff to it (+ what error messages can be expected in which situations)
* most `Alternative` combinators must be mentioned at least in passing
* parsing a language (gotta mention `identGuard` at some point)
* backtracking state + warnings or something else that uses `WriterT` + `ParsecT IO` or something (not long explanations, just something like “look what kind of thing you might want to do and here's how to do it”) (while I'm at it, there should definitely be a FAQ section at the end)
[...]

PR #50, opened 2015-10-03
Add base-4.6.0.x compatiblity

### LINKED ISSUE #46: Add base-4.6.0.x compatibility
So, why I was working on #45, I've noticed that there wasn't much missing in 4.6.0.0, except for `Data.Bool (bool)` and `Data.Either (isRight, isLeft)`.

I've prepared [a commit](https://github.com/bkaestner/megaparsec/commit/e4d1889cf2012e5ffac1edd7b1bc48b9a7baa83c) based on the 4.7.0.0 patch, that mainly provides a proof of concept to show that 4.6.0.x compatibility is feasible. However, it's a mess and not really easy to maintain.

Instead, if there's interest to add base-4.6.0.x compatibility, I would introduce a new module (`Text.Megaparsec.Comp` or similar), which re-exports the functions in base-4.7.0.0 or higher, and defines the three functions for older versions. That should give a single point of maintenance.

bkaestner:
This version introduces several `#if` for `bool` from `Data.Bool`, and (in one case), for `isLeft` and `isRight`.

Since this approach isn't really clean, I suggest to remove compatibility to base-4.6.0.0 on the second patch release of GHC 8.0, or considering an additional (single) module for those functions (if GHC 7.6 compatibility is a long term concern).

(Sorry, that took a while, I ran into some `sh` trouble due to some incompatibilities between several msys versions and couldn't install `time-*`. That's what I get for running Windows.)

This PR fixes #46.

### DIFF
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -28,7 +28,9 @@ module Text.Megaparsec.Error
   , showMessages )
 where
 
+#if MIN_VERSION_base(4,7,0)
 import Data.Bool (bool)
+#endif
 import Data.List (intercalate)
 import Data.Maybe (fromMaybe)
 
@@ -37,6 +39,11 @@ import Text.Megaparsec.Pos
 #if !MIN_VERSION_base(4,8,0)
 import Control.Applicative ((<$>))
 import Data.Foldable (foldMap)
+#endif
+#if !MIN_VERSION_base(4,7,0)
+bool :: a -> a -> Bool -> a
+bool f _ False = f
+bool _ t True  = t
 #endif
 
 -- | This data type represents parse error messages. There are three kinds
--- a/Text/Megaparsec/Prim.hs
+++ b/Text/Megaparsec/Prim.hs
@@ -40,7 +40,9 @@ module Text.Megaparsec.Prim
   , parseTest )
 where
 
+#if MIN_VERSION_base(4,7,0)
 import Data.Bool (bool)
+#endif
 import Data.Monoid
 
 import Control.Monad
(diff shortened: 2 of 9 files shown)


bkaestner:
Seems like the HPC version that's compatible to GHC 7.6. cannot get used for coveralls.

mrkkrp:
OK, I think we just won't collect data about library coverage for GHC 7.6. Please remove all tests and commits that deal with Travis CI.

bkaestner:
Removed all commits that changed the Travis configuration.

mrkkrp:
@bkaestner, Great, also remove test for `bool`, `isLeft`, and `isRight`. I'm sure the definitions are correct and there is no point is testing them.

bkaestner:
There are no dedicated tests for `bool`, `isLeft` or `isRight`. I've added them to add compatibility with GHC 7.6 in your tests, see 3078c04f8fdc70c4265e3a0e7cc46d3b6ccca023 and  a8bfdf9f569bc74e48987a4b796e20ff0d81977b. If I remove both commits, you won't be able to build your tests with base-4.6.0.x.

mrkkrp:
@bkaestner, Oh, I'm sorry, I was not attentive.

bkaestner:
No worries, the commit messages aren't that descriptive either.

ozgurakgun:
Sorry for commenting on a closed issue, but this seems like an appropriate place to ask my question: When are you guys planning to make a new release of Megaparsec that supports GHC-7.8 (and 7.6)? Thanks.

mrkkrp:
@ozgurakgun, It will be released once official tutorial is finished (we want to include it in next release). Please see #58 and also #34. I cannot give you any estimation, @neongreen writes the tutorial, not me.

ozgurakgun:
Thanks for the reply. I value good documentation and tutorials, so I understand why you would want to wait. I hope it won't take too long, and if it is going to, maybe making a point-release in the meantime may be worth it.

mrkkrp:
@ozgurakgun, Yeah, I understand that support for GHC 7.8 and GHC 7.6 will have good effect on adoption-rate of the library. On the other hand I don't want to hurry @neongreen. I think I'll release version 4.1.0 without tutorial and add it in the next version.

So, I guess Megaparsec 4.1.0 will be available by end of the week.

ozgurakgun:
That sounds like a win-win to me, great!

PR #58, opened 2015-10-11
Embedded tutorial

neongreen:
Okay, here's the beginning. (You can see it as HTML here: <http://files.artyom.me/tutorial.html>.)

### DIFF
--- a/Text/Megaparsec/Tutorial.hs
+++ b/Text/Megaparsec/Tutorial.hs
@@ -1,0 +1,465 @@
+module Text.Megaparsec.Tutorial
+(
+  -- * Introduction
+  -- $introduction
+
+  -- * Dissecting a basic parser
+  -- $dissecting
+
+  -- * Alternatives to do notation
+  -- $do-alternatives
+
+  -- * Repetition
+  -- $repetition
+
+  -- * Choice
+  -- $choice
+)
+where
+
+import Text.Megaparsec
+import Text.Megaparsec.Lexer (integer)
+import Text.Megaparsec.String (Parser)
+import Control.Applicative
+import Control.Monad
+
+{- $introduction
+
+/Parsing/ means taking text (or, say, a stream of bytes) and turning it into data; a parser is simply a function that can do that. For instance, 'read' parses a number from a string:
+
+>>> read "123" :: Int
+123
+
+Not all strings represent numbers, of course. If you try to 'read' an invalid string, you would get an error:
+
+>>> read "123xyz" :: Int
+*** Exception: Prelude.read: no parse
+
+Megaparsec is a library that helps you write parsers that are fast and produce good error messages when something goes wrong. For instance, if 'read' was a bit smarter, it could give you the following error message instead:
+
+>>> read "123xyz"
+*** Exception: line 1, column 4: expected a digit, but found 'x'
+
+Good error messages are invaluable when you are parsing something more complicated than a number. (Some examples might be: an arithmetic expression, some JSON-encoded data, or even a programming language!)
+
+To get a taste of Megaparsec, let's first write a simple parser in plain Haskell and then rewrite it using Megaparsec. Our data looks like this: @{123,456}@ (so, just 2 numbers in curly braces, separated by a comma).
+
+A parser that doesn't handle errors is pretty easy – just split the string at the comma, strip braces, and read the numbers:
+
+@
+pair :: String -> (Integer, Integer)
+pair s =
+  let (a, b) = break (== ',') s
+  in  (read (tail a), read (tail (init b)))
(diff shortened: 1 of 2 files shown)


neongreen:
(Oh, and sorry that it took so long.)

tulcod:
Would it hurt to make this literal haskell? Cause now there's code in a comment in (vacuous) code, which looks a bit weird to me.

mrkkrp:
Well, I wouldn't write tutorial at this very intoductionary level, but let it be. Mostly it's OK, could you please format text so it doesn't exceed 80 characters in width?

mrkkrp:
@neongreen, Also, since it goes in form of Megaparsec module, copy header from other modules (edit it if necessary).

neongreen:
> could you please format text so it doesn't exceed 80 characters in width?

After it's finished. Editing linewrapped text is a pain.

> copy header from other modules

Okay, sure.

mrkkrp:
@neongreen, Well, it depends on your editor, I guess. You can push here commits and we will discuss the tutorial as it progresses. When it's finished I'll merge it. Thanks for your work.

mrkkrp:
How about creating a version of this (when it's done) in markdown here: https://github.com/mrkkrp/megaparsec-site?

neongreen:
Could anybody help me with an example where limiting the scope of `try` would give a better error message? To be specific, I need a situation where:

* The full parser looks like `try a <|> b`.
* When parsing input with `a`, the error occurs at an *earlier* position than when parsing input with `b`.
* Nevertheless, `a`'s error message is more helpful to the user than `b`'s error message.

In other words, I need a situation where the “longest match” rule hurts rather than helps.

mrkkrp:
@neongreen, “more helpful” is subjective. Usually, every time when `b` gets further than `a`, we consider error message produced by `b` more “helpful”. It may be difficult to find example satisfying your criterion.

tulcod:
I'd imagine some kind of programming language where forgetting a semicolon somewhere causes the entire block of code to be parsed as one big expression.

mrkkrp:
@tulcod, in this case again, there will be no way to know if it's missing semicolon or incorrect syntax at the end of the expression. The latter error message probably would be considered more “helpful”.

tulcod:
@mrkkrp, I would disagree in the case where semantically it is clear that there should have been a semicolon. And indentation could also have indicated this. So the option "longer matches are better" can sometimes be "obviously not the choice of preference".

mrkkrp:
@tulcod, Parsers operate on syntax level, not semantic level, we cannot anticipate this level of cleverness from our parsers. From syntactic point of view, if it's possible to continue parsing with missing semicolon, then it's perfectly valid input (until parser hits incorrect input). Also, if language is indentation sensitive, incorrect indentation would produce error message even with missing semicolon.

I generally don't argue with your point of view, of course there may be situations when longer match gives worse error messages (from human point of view).

mrkkrp:
@neongreen, Don't bother to write markdown version, two versions of one document will be a nightmare to maintain. I think I will manage to convert original file copied right from GitHub repo to markdown automatically when site builds.

neongreen:
It's been a week by now and I still haven't managed to force myself to keep writing, so I'm giving up.

@mrkkrp, perhaps the “Megaparsec includes always-up-to-date tutorial in its source code, see Text.Megaparsec.Tutorial” line should be removed from the site, unless you're planning to write something yourself (and perhaps it would be easier for you, since you wouldn't be writing on an introductory level and it's always easier to write on a higher level).

I'm closing this pull request.

mrkkrp:
@neongreen, Well, there is no particular rush, you could write it for a month or two or three. Anyway thanks for your attempt. I'm currently interested in developing the library itself, so documentation will have to wait.

mrkkrp:
@neongreen, Maybe you could take everything you already have and put it on Megaparsec site in markdown with “work in progress” note. This way you might continue when you have inspiration and other people will still benefit from already written material. Open PR [here](https://github.com/mrkkrp/megaparsec-site) if you're interested.

mrkkrp:
@neongreen, Bump. I can convert what you already have to markdown and put on Megaparsec site. This may be useful for some people. What do you think?

neongreen:
@mrkkrp, I don't think it's going to be much useful in the unfinished state (pretty much all Parsec tutorials cover the basics), and if I ever want to finish it, I think not having a “semi-official” status might actually help – for instance, if I was writing it as Just Another Tutorial, I definitely wouldn't have spent literally a day trying to find justification for `<|>` not being defined as `try a <|> b` (currently it seems to me that the only justification is really “this way you don't get memory leaks”, but tell me if I'm mistaken).

PR #60, opened 2015-10-16
Improve README

bkaestner:
This commit mainly adds missing "a", "an" or "the", whenever something is used in singular, for example:

```diff
- user will see the label in error message
+ the user will see the label in the error message
```

In some cases, I've removed a word, for example "module" after `Text.Megaparsec.Char`, since it's clear you're talking about a module (otherwise "the" before the module name would be necessary).

This should provide an overall more fluid reading experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 * [Contribution](#contribution)
 * [License](#license)
 
-This is industrial-strength monadic parser combinator library. Megaparsec is
+This is an industrial-strength monadic parser combinator library. Megaparsec is
 a fork of [Parsec](https://github.com/aslatter/parsec) library originally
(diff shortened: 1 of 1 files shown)


mrkkrp:
@bkaestner, Thanks, I should admit I sometimes neglect these things. (Not sure why I would ever write “a `label ""` approach”, perhaps I initially wanted to express it differently and then edited.) Please squash everything into one commit, and we will be ready to merge.

bkaestner:
@mrkkrp Nah, the "a" was something I introduced. You wrote "prefer it to `label ""` approach". I messed the articles up in the original commit.

All commits squashed.

PR #69, opened 2015-10-30
Remove '\' from manyErr

harlanhaskins:
It gets interpreted as a tab character and this happens:

screen shot 2015-10-30 at 2 28 09 pm

### DIFF
--- a/Text/Megaparsec/Prim.hs
+++ b/Text/Megaparsec/Prim.hs
@@ -289,7 +289,7 @@ manyAcc p = ParsecT $ \s cok cerr eok _ ->
 manyErr :: a
 manyErr = error
   "Text.Megaparsec.Prim.many: combinator 'many' is applied to a parser \
-  \that accepts an empty string."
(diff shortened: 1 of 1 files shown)


mrkkrp:
Good catch. In Haskell this is however legal syntax for multi-line strings literals. Probably `CPP` extension enabling C++ preprocessor has something to do with this, because I've used multi-line literals quite a bit in other projects without any problems of this sort.

Your solution restores the word “that”, however we still have extra spaces in this case. If we don't want to use `QuasiQuotes` for such simple task, I propose we instead concat the error message from parts, like this:

```haskell
manyErr :: a
manyErr = error $ concat
  [ "Text.Megaparsec.Prim.many: combinator 'many' is applied to a parser"
  , " that accepts an empty string." ]
```

I'll think about it/fix tomorrow. Feel free to propose alternative solutions.

harlanhaskins:
I agree; that's a better solution.

PR #73, opened 2015-11-22
Canonicalise Applicative/Monad instances

hvr:
This avoids subtle performance issues and makes the code more
future proof.


mrkkrp:
Thank you. For some reason Travis is having problems with downloading of packages needed for testing (mostly it ends up with 503 server response, so I guess it's something temporary). I'll restart it/investigate the situation tomorrow.

PR #84, opened 2016-02-08
Define MonadFail & Semigroup instances

hvr:
This also enables the respective warnings flags in dev mode
to help megaparsec forward compatible.

The dependencies on `semigroup` and `fail` are conditional on
`impl(ghc >= 8)` and avoid CPP and conditionally defined instances
(which would result in an conditional API).


mrkkrp:
LGTM, Thanks!

PR #93, opened 2016-03-02
Replace `String` with `Foldable f => f Char` for some combinators

recursion-ninja:
This allows for more flexible input of any "container" holding characters rather then the specific representation of `String`. I personally use `Vector Char` often and would like to pass the vector in rather than call `toList` on the container.

Since the changed methods rely on `elem` which is an instance function of `Foldable`, it make sense to use the more general instance, just like `choice` does.

(First pull request, hope I did this right)
(Hopefully this one passes the Travis CI)

### DIFF
--- a/Text/Megaparsec/Char.hs
+++ b/Text/Megaparsec/Char.hs
@@ -54,7 +54,6 @@ where
 
 import Control.Applicative ((<|>))
 import Data.Char
-import Data.List (nub)
 import Data.Maybe (fromJust)
 
 import Text.Megaparsec.Combinator
@@ -65,6 +64,8 @@ import Text.Megaparsec.ShowToken
 
 #if !MIN_VERSION_base(4,8,0)
 import Control.Applicative ((<$>), pure)
+import Data.Foldable       (Foldable (), any, elem, notElem)
+import Prelude      hiding (any, elem, notElem)
 #endif
 
 ----------------------------------------------------------------------------
@@ -270,18 +271,13 @@ char c = satisfy (== c) <?> showToken c
 -- expecting 'E' or 'e'
 
 char' :: MonadParsec s m Char => Char -> m Char
-char' = choice . fmap char . extendi . pure
-
--- | Extends given list of characters adding uppercase version of every
--- lowercase characters and vice versa. Resulting list is guaranteed to have
--- no duplicates.
-
-extendi :: String -> String
-extendi cs = nub (cs >>= f)
-  where f c | isLower c = [c, toUpper c]
-            | isUpper c = [c, toLower c]
-            | otherwise = [c]
-
+char' c = choice [char c, char $ swapCase c]
+  where
+    swapCase x
+      | isUpper x = toLower x
+      | isLower x = toUpper x
+      | otherwise = x
+                    
 -- | This parser succeeds for any character. Returns the parsed character.
 
 anyChar :: MonadParsec s m Char => m Char
@@ -297,7 +293,7 @@ anyChar = satisfy (const True) <?> "character"
 --
 -- > digit = oneOf ['0'..'9'] <?> "digit"
 
-oneOf :: MonadParsec s m Char => String -> m Char
+oneOf :: (Foldable f, MonadParsec s m Char) => f Char -> m Char
 oneOf cs = satisfy (`elem` cs)
 
 -- | The same as 'oneOf', but case-insensitive. Returns the parsed character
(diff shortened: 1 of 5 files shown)


neongreen:
(A sidenote: when you want to update a pull request, it's okay to just push a new commit – Travis CI restarts the build when the pull request's branch changes)

recursion-ninja:
@neongreen Thanks for the pro tip!

mrkkrp:
I won't talk here about your actual commits, or why you think it's necessary to `pure` → `(:[])`. Instead I want to discuss the idea itself.

There are a number of minor problems with it:

1. Things that are unlikely to be used with the combinators would type check: sets of chars, maps to chars, etc. Excessive polymorphism without good idea behind it is a bad thing.

2. It makes it more difficult to read signatures, especially for beginners, since `String` is much more obvious than `Foldable f ⇒ f Char`.

These downsides are quite minor and would be insignificant *if* your change added some convenience, but it's hard for me to see it. Common types to represent textual data is `String` and `Text`, and the latter is not instance of `Foldable` at all AFAIK. Can you explain why you prefer to use `Vector Char` instead of `Text`? `Text` is created specially to handle… well, text.

mrkkrp:
Also see: https://www.git-scm.com/book/en/v2/Git-Basics-Undoing-Things. You need `git commit --amend`. But this is when you have one commit and you want to overwrite it. Now you need to rebase the whole banch of commits, that is, if you can explain why I should merge this.

mrkkrp:
Ah, sorry, I misread what you have changed. `noneOf` and `oneOf` could be used with sets and maps and vectors. Makes sense. I will return to this PR when I have more time. Till then try to rebase all your commits so you have one single commit with good description and title.

recursion-ninja:
I need a constant time, random access container for `Char`s (I was using `Vector`). The `Char`s in the container would also be used is a few combinators (`oneOf`,`noneOf`). It seemed a bit wasteful to fold over the container and replace the structure with a list when the underlying function didn't need the any list specific functionality (constant time prepends), only `Foldable` functionality (`elem` or `notElem`).

I like the idea of having a `Set Char`, `Vector Char`, or `Map Int Char` being an acceptable input. I do understand that having `(Int,Char)` or `Matrix Char` also being acceptable is counter-intuitive, but would a beginner really try that?

Thanks for the git related help, I'll try to make this a bit more clean tomorrow. I'm having trouble getting the changes to pass with GHC 8.x on the Travis-CI server. I keep committing the changes because I'm not sure how to run the continuous integration "locally" before a push.

Also, the `pure` -> `(:[])` was to resolve the monomorphisms restriction related to the input type of `extendi`

mrkkrp:
> Also, the pure -> (:[]) was to resolve the monomorphisms restriction related to the input type of extendi

Ah, OK. I haven't tried your PR locally, only glanced at diffs. This is likely to be merged, BTW.

recursion-ninja:
Now that the CI is green, I'll condense the git commit log later today

recursion-ninja:
I tried to `git rebase` and `squash` commits together, but it still is displaying the ugly chain of commits as I resolved the cascading unnecessary constraints. Not sure how to use `git` to make it pretty...

neongreen:
@recursion-ninja I think the easiest way to squash the commits is by doing

    $ git reset --soft b0124d1e ; the hash of the commit before your first commit
    $ git commit -m "Replace String with Foldable f => f Char for many combinators to allow for more flexible input parameters"

neongreen:
(After that, do a force-push and you should be ready to go!)

recursion-ninja:
@neongreen your `git-fu` is much stringer than mine. Thank you for teaching me to force push. Maybe one day I can become a proper ~jedi~ sith.

mrkkrp:
OK, I reviewed the PR and here are some thoughts:

1. You should not throw lots of CPP into the code to silence GHC 8 warnings about redundant constraints. If it says that they are redundant, why not just remove them? Also, please don't change how code is laid out unnecessarily, it makes it more difficult to review you changes (as [here](https://github.com/mrkkrp/megaparsec/pull/93/files#diff-e6e3e04a94bbc4e9148a82261338245aR118)).

2. I'm sorry to tell you that there is no improvement whatsoever for case-insensitive combinators. You just hide conversion to list from user, put it in `extendi` and then you again get a list. Also note that structures that are designed to be accessed by index won't do better than plain lists, because to find an element in them, they should be traversed anyway. Tree-like structures like sets and maps will do better, however. So there is no point in using `Vector Char` in this case.

To solve the problem properly, we probably should eliminate `extendi` and define case-insensitive parsers like this:

```haskell
oneOf' :: (Foldable f, MonadParsec s m Char) => f Char -> m Char
oneOf' = satisfy . f
  where f c cs | isLower c = c `elem` cs || toUpper c `elem` cs
               | isUpper c = c `elem` cs || toLower c `elem` cs
               | otherwise = c `elem` cs
```

This will be more efficient, although to see actual difference you will need quite big collection of characters. For small collections (0-10 elements), lists may be actually faster.

recursion-ninja:
I addressed point #2 in the recent commit.  I copied the `casei`, `elemi`, and `notElemi` definitions from `tests/Char.hs`. However, they are defined in both modules. I'm not sure if there is a good place for the definitions so that they can be defined once and in scope in both modules.

Regarding point #1, later tonight I'll try to remove the constraints warned about by GHC 8.0.1 entirely (no CPP) and see if the package will still compile on all 6 GHC versions `[7.6.3, 7.8.4, 7.10.1, 7.10.2, 7.10.3, 8.0.1]`.

recursion-ninja:
My git-fu was weak. I messed everything up when trying to condense the commits again...

recursion-ninja:
My git-fu was weak and I messed everything up.

How do I get the commit associated with the pull request?

mrkkrp:
What are you trying to do?

neongreen:
A pull request isn't associated with a single commit or even several commits – it's associated with a *branch* (so, a pull request basically means “I want your branch to look more like mine”). In this case, the original branch is `mrkkrp:master`, and your branch is `recursion-ninja:master`.

If you messed up locally and haven't done any pushes yet, the simplest way to fix it is to do

    git clone https://github.com/recursion-ninja/megaparsec

neongreen:
(I mean, remove the folder you have locally and then do the command.)

recursion-ninja:
I got the git repo fixed.

Regarding @mrkkrp's  point 1: 

I removing the CPP related to the type constraints and the package built fine on all the GHC versions *except* version `8.0.1`. This one compiled with many warnings about unnecessary type constraints. The `-Werror` made this compilation fail for that compiler version.

I then removed all the type constraints for that had generated the warnings. This compiled on GHC `8.0.1` *only* and failed to type check on all the lower GHC versions. 

I'm not sure how to proceed with the CPP related to the type constraints. I'm actually rather surprised the changes `String` -> `Foldable f => f Char` would affect any of the constraints which were generating warnings. 

I think @mrkkrp's point 2 has been addressed. Nice catch that `extedni` wasn't necessary any more!

mrkkrp:
OK, looks good. I will return to this later to investigate that CPP issue.

mrkkrp:
Applied as dccfc82.

Issue #70, opened 2015-11-04
token function to receive currentPos instead of nextPos


chamini2:
I know the `token` function receives a function with signature `Int -> SourcePos -> t -> SourcePos` to calculate the position of the next token, however, in my particular case, I can't give such a function, since I don't know how many spaces there may be between tokens.

Is there any chance to get a `token` that instead receives a function that gives the `SourcePos` of the **current** token?

I suppose that you ask for `nextPos` for design purposes, and it may be a hard change, or even impossible with the current state of the library, but I wanted to know if it's possible.

mrkkrp:
Please give realistic example when this would be useful. If it's a god idea, we could add one more version of `token` function.

chamini2:
Well, I have a parser on a `data` that's been already parsed using `alex` and `happy`, this tokens store the location span information (i.e. where in the file it starts and where it ends), but in the lexer (`alex`), I just ignore whitespace, so an expression with the form

    x=2

parses the same as

    x       =        
                 2

Both of this give me the same information to be parsed, the only different information is source location, working with tokens like this, there's no way to infer the next token's position, as it could be at the end of this token (like in the first example), or on the next line (like in the second example).

For clarification, my current `token` function usage looks a bit like this:

    class Span a where
            getSpan :: a -> SrcSpan

    satisfy' :: Span x => (x -> Either [Message] x) -> Parsec [x] x
    satisfy' guard = token nextPos guard
        where
            nextPos :: Int -> SourcePos -> x -> SourcePos
            nextPos _ _ x = spanToPos (getSpan x)

`spanToPos` gets the ending position of the `SourceSpan`.

So I'm just ignoring the last `SourcePos`, since there's no information it can provide.

This means that all my error messages are positioned a few spaces before than it should.

mrkkrp:
I've not used `alex`, so let's clarify your situation a bit. You run `alex` on your input and it produces collection of tokens, every token has associated position (where it starts in input file and where it ends) and its somewhat normalized form where `x = 2` is the same as `x=2` (I mean you still have it as textual fragment, right?). Now you want to parse this fragment using Megaparsec.

Please correct me if I'm wrong.

mrkkrp:
I understand what you mean, there will be new version of `token` in Megaparsec 4.3.0 tailored exactly for this approach to parsing. Thanks for the report.

mrkkrp:
@chamini2, There is something for this on master now. Can you try it and give some feedback?

[...]

PR #94, opened 2016-03-24
More flexible position-advancing function

mrkkrp:
This PR aims to improve experience of users who work with custom input streams. In particular with such streams where tokens have start/end positions attached (for example when input stream is generated by alex/happy).

TODO list:

- [x] Check how this affects performance
- [x] Update `CHANGELOG.md`
- [x] Don't forget to add `@since` to new functions
- [x] Write test that checks that `string` never consumes input when it fails
- [x] Write tests for the new feature and also test non-`String` streams (`Text`, `ByteString`)


mrkkrp:
@chamini2, Please see if this PR solves the problem you describe in #70.

Issue #83, opened 2016-02-08
Q: Recommended way to handle source-locations for languages which support include files?


hvr:
I have been wondering what the recommended way to handle languages with include files is.

I currently have a stream of tokens I pass to the parser (I have a separate CPP pass, which I want to keep that way to keep the `ParsecT` stack `IO`-less), with a special token to denote file changes, i.e. 

    data Token = ... | SourcePush FilePath | SourcePop | ...`

With Parsec I would have maintained a `StateT`  stack of `SourcePos`itions updated everytime the respective `Token` came along. However, since with `megaparsec` the user-state is gone, I'm now wondering if there's a different/better/recommended way to handle this case without having to add a `StateT` around `ParsecT` to emulate Parsec's custom state.

PS: would it be conceivable if megaparsec came with builtin support for maintaining a `SourcePos` stack? This would probably simplify handling include files a lot, unless I'm missing something.

PS2: what if handling  `SourcePos` updating would move away from the `token` primitive into the  `Stream` class? Why is `SourcePos` updating a concern of the `token` function?

mrkkrp:
For this you definitely need custom backtracking state and putting `ParsecT` inside `StateT` is the only option that comes to mind. Can you elaborate why exactly you consider it not the best solution? Everything in Megaparsec works with any instance of `MonadParsec`, so I don't see problems with this approach, should be just as easy as Parsec's user state.

----

> PS: would it be conceivable if megaparsec came with builtin support for maintaining a SourcePos stack? This would probably simplify handling include files a lot, unless I'm missing something.

This is an interesting idea, although I don't see how it can be done in a backward-compatible way, because constructor of `State` is public (not a bad thing per se, because we support incremental parsing now and construction/manipulation of `State` may be of interest for users).

Compatibility concerns aside, we could easily have `[SourcePos]` instead of `SourcePos` in `State` and then switching contexts would be as trivial as pushing to and popping from the list. I see no way it can affect performance either.

----

> PS2: what if handling SourcePos updating would move away from the token primitive into the Stream class? Why is SourcePos updating a concern of the token function?

It's Parsec's legacy. Somehow it didn't come to my mind when I reorganized things when I started Megaparsec. I love the idea, but again it will break backward-compatibility.

Thanks for the input, I think these changes/features will be available in Megaparsec 5.0.0, which is going to be the next release after upcoming Megaparsec 4.4.0.

mrkkrp:
[...]

Issue #98, opened 2016-03-31
Missing instance of Read for ParseError (or similar functionality).


expipiplus1:
It would be useful for testing parsers to be able to read an expected `ParseError` from a file and compare to the result of running a parser. At the moment the only (simple) way to do this is to compare the parse errors as strings, which is a little gross.

It's also a little odd that the instance for `Show` for `ParseError` pretty prints the error, rather than doing what `derive Show` would do. I presume that this is due to historical reasons?

mrkkrp:
We cannot allow construction of `ParseError` in any way other than via smart constructors from `Text.Megaparsec.Error` module. At least for now, because we need to keep `[Messages]` sorted to properly handle comparison of `ParseError`s. So, currently derived `Read` instance would open the possibility to construct invalid `ParseError`s. I'm thinking of using `Set` instead of `[Message]` in Megaparsec 5.0.0 (need to check how this affects performance). If we swich to `Set`, then there will be no way to constuct “invalid” `ParseError` even if we export `ParseError` constructor and derive `Read`. I also like the idea of deriving `Show` and introducing `parseErrorPretty` or similar function for pretty-printing.

expipiplus1:
Did you mean `showErrorPretty` in the last line there. If so hang onto `parseErrorPretty` it'd be a good solution to reading `ParseError`s from strings.

mrkkrp:
`parseErrorPretty` is meant for pretty printing (i.e. it's name of type plus the word “pretty”), but it's not that important how it's called.

----

Although to derive `Read` instance of `ParseError` we need to derive `Read` of `SourcePos` too, that may be tricky, since line and column numbers must be greater than 1 — again we should only allow construction the `SourcePos` via smart constructors because of this.

Derived `Show` instances for these things will probably be in Megaparsec 5.0.0, but I'm not sure about `Read` instances.

expipiplus1:
In case anyone runs into the same issue, here's a set of `newtype` wrappers to read and show `ParseError` and friends.

``` haskell
import           GHC.Read                        (Read (..), expectP)
import qualified Text.Megaparsec.Error           as M
import qualified Text.Megaparsec.Pos             as M
import           Text.ParserCombinators.ReadPrec (choice, prec, step)
import           Text.Read                       (parens)
import           Text.Read.Lex                   (Lexeme (Ident))

--------------------------------------------------------------------------------
-- Workaround for https://github.com/mrkkrp/megaparsec/issues/98
[...]

Issue #97, opened 2016-03-30
Support for custom datatypes as parse errors


wdanilo:
Hello!
First of all congratulations! I've been watching the state of the art of parsing in Haskell for long time and I was hoping trifecta will go the way megaparsec went. In fact I was planning to create my fork of Parsec to support the "recovery" feature and something that is yet missing from megaprasec - namely the ability to raise custom datatype-based errors, not limiting ourselves to strings only. Do you plan to add support for it in the future?

mrkkrp:
Thanks! It's an interesting idea, do you have some specific design of how this can be added to the library?

wdanilo:
It should be just added as a type parameter denoting what datatype I want to use when reporting errors. This datatype should implement some kind of a standard parser error type class that will handle all possible errors that can happen internally, like unexpected token or whatever. This way you can just `runParser` and get `Either MyErrorType Ok`. Does it clarify it to you?

mrkkrp:
I see. There are several ways to add this functionality, so I just wanted to hear about the option that you have in mind. I do think it's a good feature and it will probably be in upcoming Megaparsec 5.0.0.

wdanilo:
Glad to hear that! If you will have any specific questions regarding implementation / ideas - contact me whenever you'll need to! :)

mrkkrp:
A couple of things in your approach I don't like. First, AFAICT, this will add one more type variable to `ParsecT` type, and thus will make signatures more verbose — I would like to avoid that. Another thing is that returned type should somehow represent “normal” exceptions (unexpected, expected stuff) and in addition to that we would like to pass additional, non-string info through parse errors. Type classes can solve this, but implementer of new error type will need to do work that is already done in `Text.Megaparsec.Error` in order to get get it right. Given such API I would personally defined something like:

```haskell
data CustomError = CustomError ParseError MyData
```

And then I would make it instance of type class that implements things related to parse errors. I would delegate most of work to internal `ParseError` where all things are already handled well out-of-the-box.

Another thing is that to use this feature, we need to essentially allow failing with already pre-constructed errors (additional work, additional place to make mistakes, for example there is nothing to protect you from giving it wrong position). This may be the price for additional power, but I like current approach where library takes care of error positions and you only tell it which messages to include.

----

[...]

PR #99, opened 2016-04-10
Position and error reform

mrkkrp:
This PR will track progress of “Position and error reform” that should hopefully resolve issues #83, #97, and #98.

Road map:

- [x] Change definition of `SourcePos`. Provide `Show` and `Read` instances for it. Represent `sourceLine` and `sourceColumn` as `newtype` allowing construction of only positive numbers. Use `MonadThrow` to reject non-positive numbers. Provide primitives that would allow pure (non-monadic), fast incrementation of these things and converting them to `Word`. Provide separate function for pretty-printing of `SourcePos`. Re-think API of `Text.Megaparsec.Pos` module, think what primitives we should put there. Adjust test suite.

- [x] Use new `Text.Megaparsec.Pos` to reform `Text.Megaparsec.Error`. Provide `Show` and `Read` instances for `ParseError`. Instead of flat `SourcePos`, use `NonEmpty SourcePos` to support include files. Instead of `[Message]` use `Set Message`. Change in performance will be minimal, because although list is faster than `Set` for small numbers of elements, `[Message]` should be kept sorted, and work that is necessary to keep it sorted sort-of makes performance of entire thing close to performance of `Set` even for small numbers of elements. `Set` also scales better and it's just the right thing to use here. `errorMessages` should return set, so we can avoid conversion to list for hints (hints should be `Set` based too under the hood). Union of sets is faster than merging of two lists (keeping result sorted), so actually we might get better performance for `<|>` here (check it). Provide separate function for pretty-printing of `ParseError`. Adjust test suite.

- [x] Decide how to give users convenient way to work with include files. `ParseError` and `State` should already support stacks of positions, but what is the best way to add ability to switch files to API? Add new tests for the new feature.

- [x] Decide how to allow failing with custom datatype as parse error. See discussion in #97.


chamini2:
Hey, I was considering that maybe since `setPosition` and `getPosition` now will handle `NonEmpty SorucePos`, maybe you could include a `setCurrentPosition` and `getCurrentPosition` that just handle the top position, not the whole stack.

Or if you want to keep backwards compatibility, you could use the original names for handling the top position only and include two new names `setPositionStack` and `getPositionStack`.

mrkkrp:
I have made `getPosition` and `setPosition` work with only top-level position, because I think it will be most common use case. Users can always get and set the whole stack with help of `getParserState` and `updateParserState` if necessary.

I have also introduced `pushPosition` and `popPosition` like you have proposed in #83.

That's about it. I will need to review everything, update change log, and this will be merged. Quite a few exiting features and closed issues :-)

PR #100, opened 2016-04-17
Update position before failure of `pEof`

chamini2:
Tested the functionality introduced in #94, and noticed that it didn't work for cases where the `eof` parser failed. 
Adding this fix to report the error in the updated position instead of the old one seems to solve the problem.

@mrkkrp, let me know if this makes sense or not.

### DIFF
--- a/Text/Megaparsec/Prim.hs
+++ b/Text/Megaparsec/Prim.hs
@@ -654,11 +654,13 @@ pWithRecovery r p = ParsecT $ \s cok cerr eok eerr ->
   in unParser p s cok mcerr eok meerr
 {-# INLINE pWithRecovery #-}
 
-pEof :: Stream s t => ParsecT s m ()
-pEof = label eoi $ ParsecT $ \s@(State input pos _) _ _ eok eerr ->
+pEof :: forall s m t. Stream s t => ParsecT s m ()
+pEof = label eoi $ ParsecT $ \s@(State input pos w) _ _ eok eerr ->
   case uncons input of
     Nothing    -> eok () s mempty
-    Just (x,_) -> eerr (unexpectedErr (showToken x) pos) s
+    Just (x,_) ->
+      let !apos = fst (updatePos (Proxy :: Proxy s) w pos x)
+      in eerr (unexpectedErr (showToken c) apos) s
 {-# INLINE pEof #-}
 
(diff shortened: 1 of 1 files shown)


mrkkrp:
Good catch. In the rest, does it work well? Thank you for your contribution so far, I really appreciate it. I would like to ask you to write shortish introduction how to use Megaparsec with happy/alex, because I'm not familiar with these tools. Nothing sophisticated is needed, just a short basic example.

If you decide to help, you can just open a PR here:

https://github.com/mrkkrp/megaparsec-site/

You can use current master, I will make sure that upcoming changes are reflected in the new tutorial (if you decide to write it).

----

About this problem, let's do the following: I'm working on rather fundamental changes to the library right now, so considerable part of code base is being rewritten. I can merge your PR, but chances are this part of code will be changed anyway. Also, since you have discovered such a flaw, we need a test for it. I will keep this PR as reminder for me to fix this and add new test when code “stabilizes”.

chamini2:
I would gladly write the tutorial / introduction. But, if what you really want is to show a use case in which the tokens *know* their position, I could write one that uses megaparsec only if you like.

My use case was with Alex/Happy because I knew that parsing option better at the moment. But I would have needed the same features even if I used megaparsec for all the parsing.

I had to parse an input file and generate a new grammar based on some of the input, then I had to parse certain parts of the file with the generated grammar, that's when megaparsec came in handy. I think a good tutorial is a minimized version of what I did. An input that has expressions and operator definitions to read the expressions correctly and output it in function application (Haskell-style).

For example:
```
infixL + - 4
infixL * / 5
infixR ^   6

1 + 2 * 3 - 4
```
Output:
```
(-) ((+) 1 ((*) 2 3)) 4
```

This can be written using only megaparsec or combining Alex/Happy and megaparsec. What do you think? Should I do it combining them or using only megaparsec?

mrkkrp:
Thank you. I think Alex/Happy are popular and considered a good option for serious tasks (AFAIK Haskell's parser itself is implemented with them). So, a tutorial involving these tools would be more valuable as it would show how to use them with Megaparsec and I guess it will be obvious anyway how to use the new position-advancing machinery even without Alex and Happy.

mrkkrp:
Closing the PR, you can track progress of this issue in #102 instead.

PR #149, opened 2016-10-10
Remove un-needed unsafeCoerce

erikd:
`Pos` is a newtype wrapper around `Word` and the `unsafePos` function
is marked `INLINE`, so `unsafeCoerce` is complete overkill.


mrkkrp:
Yeah, the definitions are equivalent. I have nothing against it — it's more readable this way perhaps.

erikd:
Thanks @mrkkrp !

PR #161, opened 2016-12-27
Make indentBlock/IndentNone work with many like IndentMany/Some does

liskin:
Even though the haddock for `Text.Megaparsec.Lexer.space` says that

    Parsing of white space is an important part of any parser. We
    propose a convention where every lexeme parser assumes no spaces
    before the lexeme and consumes all spaces after the lexeme;

all the indentation-sensitive parsing combinators assume/consume
whitespace _before_ the thing to be parsed. This would normally mean
they can't be used with combinators like `many` and `some` without using
`try` (and sacrificing performance). Fortunately `indentBlock` also
consumes whitespace _after_, but unfortunately it didn't do that in the
`IndentNone` case. Now it does and it works with many and some without
try!  \o/

### DIFF
--- a/Text/Megaparsec/Lexer.hs
+++ b/Text/Megaparsec/Lexer.hs
@@ -289,7 +289,7 @@ indentBlock sc r = do
   ref <- indentLevel
   a   <- r
   case a of
-    IndentNone x -> return x
+    IndentNone x -> sc *> return x
     IndentMany indent f p -> do
       mlvl <- optional . try $ C.eol *> indentGuard sc GT ref
       case mlvl of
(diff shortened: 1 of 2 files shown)


liskin:
With lts-7 and megaparsec-5.0.1, this seems to be a workaround:

    indentBlockNone :: Parser a -> Parser a
    indentBlockNone p = L.indentBlock scn $
        L.IndentNone <$> p <* scn
        -- the "<* scn" belongs to L.indentBlock instead :-(

Unfortunately the haddock for `indentBlock` explicitly forbids the use of newline-consuming parsers, but if the tutorial example can do it (https://github.com/mrkkrp/megaparsec-site/commit/06e77e2b32d7b46beec67edd895846306db657a3#diff-c7c22e321d07b54edf6fa1ed7dc7b75dR36), it's probably not that harmful. :-)

mrkkrp:
Good catch, thanks!

Issue #101, opened 2016-04-18
Number lexer


minad:
Hi @mrkkrp,

concerning the number lexer, I would like to have a few things :)
Would you accept patches concerning that?
1. Parsing arbitrary prefixes and bases. Your `nump`goes in that direction, however it is restricted too much by haskell syntax. What about a combinator like `numberBase base`? The base can be anything up to 16. The prefix can then be parsed separately with `string'`
2. Supporting `_` in the number strings, e.g., 3.141_592_653_596, 1_000_000_000

While 1. is a simple addition, I don't know how one could add 2. Maybe by passing an additional digit parser?

Daniel

mrkkrp:
We do not currently parse prefixes as part of parsers that are shipped with the library, [see here](https://hackage.haskell.org/package/megaparsec-4.4.0/docs/Text-Megaparsec-Lexer.html#v:hexadecimal). Parsing of numbers with arbitrary base may be added, there is nothing wrong with that idea.

Your second proposition looks like rather a custom thing, for which you can write your own parser that accepts underscores. I don't think `1_000_000` notation is used frequently (in fact, I see this sort of notation for the first time; I think spaces are used much more frequently as delimiters). Also, I've seen something like this `1 0000 0000`, so it's a custom thing that should be writen by user for concrete use-case, IMO.

minad:
Yes, you don't. However you use read and add the prefix for the parsing step internally. I meant we could switch to a more general string to integer converter. This would allow the addition of the aforementioned parser.

Actually, the underscore is not that uncommon. See for example http://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html, http://mlton.org/SuccessorML and http://docs.ruby-lang.org/en/2.0.0/syntax/literals_rdoc.html#label-Numbers

 Yes, I am doing this with my own parser right now. However since the whole point of splitting up this Parsec.Token module was to make it more generic, I thought I bring it up ;)

mrkkrp:
No worries, thanks for the proposition!

mrkkrp:
Feel free to open a PR implementing 1 if you want this feature, but I won't work on this myself (there are other more important things to work on in Megaparsec).

mrkkrp:
[...]

PR #181, opened 2017-01-30
add integerBase

minad:
See https://github.com/mrkkrp/megaparsec/issues/101

This still needs tests and the char functions should probably go to Megaparsec.Char.

Is this generally ok?

### DIFF
--- a/Text/Megaparsec/Lexer.hs
+++ b/Text/Megaparsec/Lexer.hs
@@ -42,6 +42,7 @@ module Text.Megaparsec.Lexer
     -- * Character and string literals
   , charLiteral
     -- * Numbers
+  , integerBase
   , integer
   , decimal
   , hexadecimal
@@ -52,13 +53,14 @@ module Text.Megaparsec.Lexer
   , signed )
 where
 
-import Control.Applicative ((<|>), some, optional)
+import Control.Applicative ((<|>), some, optional, empty, many)
 import Control.Monad (void)
-import Data.Char (readLitChar)
 import Data.List.NonEmpty (NonEmpty (..))
 import Data.Maybe (listToMaybe, fromMaybe, isJust)
 import Data.Scientific (Scientific, toRealFloat)
+import Data.Foldable (foldl')
 import qualified Data.Set as E
+import qualified Data.Char as Char
 
 import Text.Megaparsec.Combinator
 import Text.Megaparsec.Error
@@ -189,8 +191,7 @@ skipBlockCommentNested start end = p >> void (manyTill e n)
 -- Indentation
 
 -- | Return current indentation level.
---
--- The function is a simple shortcut defined as:
+---- The function is a simple shortcut defined as:
 --
 -- > indentLevel = sourceColumn <$> getPosition
 --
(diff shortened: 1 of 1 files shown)


mrkkrp:
Thanks. Here are some points:

* Just ignoring a character (for example `,` in `1,000,000`) is not good enough. When characters delimiters are used, they are used in a meaningful way, so we should parse them in a meaningful way as well. I.e. accepting something like `,,,,0` along with `1,000,000` is not OK.

* I think we should not combine the feature to support `1_000_000` and `1,000,000` things with parser for numbers in given base. The two features seem to be completely independent and never used together. How often would anyone want to parse a number in base 20 and delimit characters with `,` at the same time?

* If we start to mess with “delimiting characters”, the feature should work not only for integers, e.g. this is a valid use case: `1,000.00`.

* Please be careful not to break existing formatting as on lines 192-194, or I'll need to cherry-pick/amend your commits to fix it back — not fun!

* I think the character parsing functions should **not** go to `Text.Megaprasec.Char`, as their use outside of `integerBase` seems to be non-existent.

* Could something like [this](https://hackage.haskell.org/package/base-4.9.1.0/docs/Numeric.html#v:readInt) be used instead of manual calculations? Also, please don't covert existing code unless we previously explicitly agreed that it's OK.

(Note to self: these `readDec`, `readOct`, and `readHex` friends could actually be better than prefixing numbers and feeding them into `read`.)

OK, sorry for making so many remarks! I appreciate you're working on this, I just need to keep things in good shape :-) Bottom line: I think it's best to concentrate just on a parser that parses numbers in arbitrary base for now.

minad:
1. The ignore parser does exactly that. It requires the ignored character to be preceded and followed by a digit.
2. How then? integerBase should just function as a generalized integer parser with generalized base and ignored delimiters.
3. Yes, we could also provide such a thing for floats.
4. Typo.
5. Ok with me.
6. I've seen this function, but it seemed to be too involved due to the additional isDigit checking. However now that I looked again, it seems readInt uses a optimized internal parser, which would make it useful (See https://hackage.haskell.org/package/base-4.9.1.0/docs/src/Text.Read.Lex.html#valInteger). I strongly disagree with the way you handle the parsing now using "0x" prefixing. Therefore I changed it.

I appreciate that you want to keep things in good shape.

minad:
Maybe I should add my use case: Assume I want to parse numbers of the format 2'1000_0111_0000, 10'123_456_789 etc

How could I do that using the given infrastructure. This PR tries to provide the necessary combinator.

mrkkrp:
I don't see how to add the feature with delimiter characters nicely without breaking compatibility, perhaps it is a bit too specific, I'm not sure. I don't like the prefixing technique too (although it's totally safe in our case), I should find time to benchmark the parsers to find out if switching to `readDec` and friends can improve the situation. These reader functions may perform better than gerenal `read` we use right now.

If you build `integerBase` on top of `readInt` from `Numeric`, we can merge that. If you do, don't forget to add `@since 5.2.0` note, changelog note, and tests.

minad:
What is the problem with backward compatibility? decimal, hexadecimal etc could stay as they are.
If you don't want to add the ignore feature, it is also ok for me.

mrkkrp:
Because ideally this should be an argument for already existing parsers that produce integers, floats, scientific. We don't want to add a bunch of mostly identical combinators, it would be ugly, but at the same time we can't add new arguments to existing parsers because it would break backward compatibility.

minad:
Ok, probably it is too much and would be used only seldom.

mrkkrp:
Are you interested in `integerBase` parser without ignoring characters?

mrkkrp:
I want to replace implementation of `decimal` and friends, I could add `integerBase` as well. Or is it only of interested when combined for the "ignorable characters" feature?

minad:
I need the ignore feature. This is somehow all or nothing since if megaparsec doesn't support it directly I have to use my own combinator and could not reuse integerBase. But I am unsure if this feature is out of scope for megaparsec. It doesn't make much sense to add weird parsers which are never used.

PR #216, opened 2017-06-14
Move QuickCheck dep out of library

fosskers:
I was checking out the dep graph of a library of mine which uses Megaparsec:

deps

Seeing `QuickCheck` in there as a build-dep for the library struck me as odd. I figured it might be `Arbitrary` instances, and it was, so I did this quick PR to move them out (as orphans) into the test suite. This reduces the dep burden on people who use megaparsec, as there are some QC-only transitive deps that are pulled in by it.

The module I added might not have the best name. Would `Instances` or `Orphans` be better?

### DIFF
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -41,7 +41,6 @@ import Data.Set (Set)
 import Data.Typeable (Typeable)
 import GHC.Generics
 import Prelude hiding (concat)
-import Test.QuickCheck hiding (label)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Set           as E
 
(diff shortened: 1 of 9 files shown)


fosskers:
The new dep graph:

deps

mrkkrp:
`Arbitrary` instances were added on purpose because I was tired to define the same orphans in all test suites again and again. I still think it was a good decision overall. I'm not sure about `QuickCheck`, but Megaparsec 6 will probably drop `exceptions`.

fosskers:
As in, test suites in other libraries that use `megaparsec` and also rely on the `Arbitrary` instances of its internal types?

mrkkrp:
> As in, test suites in other libraries that use megaparsec and also rely on the Arbitrary instances of its internal types?

Yes, exactly. At least those that I authored.

> Wow those took a long time. Looks like you're not using stack and caching deps between CI runs?

Caching yes, stack no. I believe in Cabal on CI so the code is tested in more "adverse" conditions (i.e. Cabal-only users (if they exist) still can build it). Before caching it took even longer. GHC has hard time compiling the test suite if I remember correctly, actually compiling it takes longer than running it.

mrkkrp:
Is this dependency a deal-breaker for you for some reason or you just would like as lightweight library as possible?

fosskers:
It's not a deal-breaker, no, but it does feel like a wart. Anyone who depends on `megaparsec` inherits its "larger than necessary" dependency DAG, which costs them time during dep resolution / building. With `QuickCheck` right in the library, Megaparsec devs are saved time from having to manage the `Arbitrary` instances, at the cost of down-stream user time and binary sizes.

I'd say in general I strive for the smallest dependency DAG possible, but that's me.

mrkkrp:
I have to close this one for now as it conflicts with my another PR, but I'll keep it mind that making dependencies more lightweight is desirable for Megaparsec's users. When Megaprasec 6 is closer to its release, I'll revisit this.

fosskers:
Thanks.

PR #225, opened 2017-06-29
Add a MonadFix ParsecT instance

feuerbach:
Example usage:

``` haskell
{-# LANGUAGE RecursiveDo, TypeApplications #-}
import Text.Megaparsec
import Text.Megaparsec.Char
import Control.Monad.Fix
import Data.Void

withRange
  :: (MonadParsec e s m, MonadFix m)
  => ((SourcePos,SourcePos) -> m a)
  -> m a
withRange f = do
  Just pos1 <- getNextTokenPosition
  rec
    r <- f (pos1, pos2)
    pos2 <- getPosition
  return r

main = do
  parseTest @Void (withRange $ \pp -> (,) pp <$> string "ab") "abcd"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@
 * More lightweight dependency tree, dropped `exceptions` and `QuickCheck`
   dependencies.
 
+* Added a `MonadFix` instance for `ParsecT`
+
 ## Megaparsec 5.3.1
 
 * Various updates to the docs.
--- a/Text/Megaparsec.hs
+++ b/Text/Megaparsec.hs
@@ -400,6 +400,14 @@ pPlus m n = ParsecT $ \s cok cerr eok eerr ->
   in unParser m s cok cerr eok meerr
 {-# INLINE pPlus #-}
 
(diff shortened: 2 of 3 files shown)


mrkkrp:
Can you add a test for this?

Am I correct that it'll blow up if the inner parser of `mfix` fails? (Remembering similar magic about `MonadFix` for `Maybe`, I'm somewhat in doubt if `error` is reachable. At least from looking at the code, it looks like it's reachable. Have no time to try right now.)

feuerbach:
Yes, tests are a good idea. (There are laws for MonadFix, so I should test those.)

`error` is reachable if the error message depends on the value:

```
parseTest @Void (mfix $ \x -> fail x) ""
```

But this is fine:

```
parseTest (mfix $ \x -> fail "xxx" :: Parsec Void String ()) ""
```

feuerbach:
On a second thought, those laws aren't particularly amenable to testing, so I went with the example.

mrkkrp:
Please also add a changelog entry. (Use `[skip ci]` in the commit message, or Travis will take forever again.)

feuerbach:
Done.

Issue #205, opened 2017-05-05
Implement optional displaying of releavant region of input in error messages


mrkkrp:
For version 6, I want to have something like this:

```
my-file.txt:8:6:

x = 1a
     ^

unexpected 'a'
expecting newline or the rest of integer
```

Regarding implementation, I think we are not going to store input regions for this, so no changes to the parsing process should be necessary (as we don't want to make things slower, in fact version 6 is about making things faster). The most obvious thing (and robust) is to traverse original input updating position till we find the position from `ParseError`. Remembering where the most recent line begins will allow us to extract just the relevant line of input and display it.

I'm not sure how to do it better without making the feature fragile. It seems that we can't use total number of processed tokens to figure out where `ParseError` happened, because in some cases, for example when using `region` to overwrite error position, total number of processed tokens would be out-of-sync.

This feature should be optional.

mrkkrp:
@recursion-ninja, any progress with this one?

recursion-ninja:
@mrkkrp my weekend got away from me, so no there's no substantial progress. However, I have given it a fair bit of thought.

I'm not exactly sure how to implement it on an arbitrary stream of tokens. We would need a way to determine the relevant scope for a given parse error. It will be much easier to implement if constrained on `Token s ~ Char` so we can determine the beginning and end of a given line. 

There might be a way to generalize this for an arbitrary stream of tokens by passing in a "scoping" function as a parameter which signals when a new "scope" for a parse error in the stream has been consumed. For streams where `Token s ~ Char` this function would signal a new "scope" (line) has been entered when a `'\n'` or a `\r\n` has been consumed.

mrkkrp:
Well, let's see. We are going to need functions from the [`Text.Megaparsec.Stream`](https://github.com/mrkkrp/megaparsec/blob/master/Text/Megaparsec/Stream.hs) module. Note that they work for any `Stream` and to consume some input, it must be an instance of `Stream`, so the methods of the class seem to be the right tool.

I think to print the relevant line, we are going to take the whole input as an argument. Since the feature should be optional, maybe it's best to have a separate function for this (similar to `parseErrorPretty`):

```haskell
parseErrorPretty'
  :: ( Ord (Token s)
     , ShowToken (Token s)
     , ShowErrorComponent e
     , Stream s )
  => s
  -> ParseError t e
  -> String
parseErrorPretty' = undefined
```

[...]

PR #233, opened 2017-07-11
Rendering parse error context #205

recursion-ninja:
An *initial* implementation to render the line of a parse error along with the parse error. Addresses issue #205.

The bar on the side is intended to mimic GHC's 8.2.0's new contextual warning/error rendering.

Let me know your feedback .

Some example renderings:
```
*ghci> let parseTest2 = (\p s -> either (putStr . parseErrorPrettyWithLine s) print $ parse p "example.txt" s) :: Show a => Parsec Void String a -> String -> IO ()
*ghci> parseTest2 (count 10 numberChar) "123456789"
example.txt:1:10:
   |
 1 | 123456789
   |          ^
unexpected end of input
expecting numeric character
*ghci> parseTest2 (count 10 (numberChar <|> Text.Megaparsec.Char.newline)) "12\n3\n45y"
example.txt:3:3:
   |
 3 | 45y
   |   ^
unexpected 'y'
expecting newline or numeric character
*ghci> parseTest2 (count 10 (numberChar <|> Text.Megaparsec.Char.newline)) "12\n3\n45\n"
example.txt:4:1:
   |
 4 | <empty line>
   | ^
unexpected end of input
expecting newline or numeric character
*ghci> putStrLn "Note the contextual line number padding on the left-hand side:"
Note the contextual line number padding on the left-hand side:
*ghci> parseTest2 (count 10010 (numberChar<|>Text.Megaparsec.Char.newline)) $ replicate 10000 '\n' <> "123456789"
example.txt:10001:10:
       |
 10001 | 123456789
       |          ^
unexpected end of input
expecting newline or numeric character
```

### DIFF
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -15,12 +15,13 @@
 -- You probably do not want to import this module directly because
 -- "Text.Megaparsec" re-exports it anyway.
 
-{-# LANGUAGE CPP                #-}
-{-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE DeriveFunctor      #-}
-{-# LANGUAGE DeriveGeneric      #-}
-{-# LANGUAGE FlexibleContexts   #-}
-{-# LANGUAGE FlexibleInstances  #-}
+{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE DeriveDataTypeable  #-}
+{-# LANGUAGE DeriveFunctor       #-}
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Text.Megaparsec.Error
   ( ErrorItem (..)
@@ -277,24 +278,28 @@ parseErrorPrettyWithLine stream e = mconcat messageComponents
 -- | Renders the source context to improve error message quality.
 
 renderLineMessage
-  :: ( Enum (Token s)
+  :: forall s .
+     ( Enum (Token s)
      , Stream s )
   => s -> SourcePos -> String
 renderLineMessage s p = unlines [ paddingLine, contextLine, pointingLine ]
   where
-    paddingLine   = padding <> "|\n"
-    contextLine   = " " <> lineNumberStr <> "| " <> cs
-    pointingLine  = padding <> "| " <> (replicate columnNum ' ') <> "^"
+    paddingLine   = padding <> "|"
+    contextLine   = " " <> lineNumberStr <> " | " <> cs
+    pointingLine  = padding <> "|" <> replicate columnNum ' ' <> "^"
 
     lineNumberStr = show $ unPos linePos
     padLength = length lineNumberStr + 2
     padding = replicate padLength ' '
 
-    cs = chr . fromEnum <$> ts
+    cs =
+      case ts of
+        [] -> "<empty line>"
+        xs -> chr . fromEnum <$> xs
     ts = chunkToTokens (Proxy :: Proxy s) lineVal
     lineVal = streamIter grabLine linePos s
     linePos = sourceLine p
-    columnNum = unPos (sourceColumn p) - 1
+    columnNum = unPos $ sourceColumn p
 
 -- | Applies a "taking" function @n@ number of times to a 'Stream' and returns
 -- the last result of the "tking" function, discarding previously taken chunks.
@@ -306,9 +311,9 @@ streamIter
   -> Pos                  -- ^ Number of times to apply the function.
   -> s                    -- ^ Stream to be taken from
   -> Tokens s             -- ^ The the last segment taken from the stream
-streamIter f p ts = fst $ go (n-1) (f ts)
+streamIter f p ts = fst $ go (x-1) (f ts)
   where
-    n = unPos p
+    x = unPos p
     go 0 val    = val
     go n (_,xs) = let val = f xs
                   in go (n-1) val
@@ -321,16 +326,13 @@ streamIter f p ts = fst $ go (n-1) (f ts)
 grabLine :: (Enum (Token s), Stream s) => s -> (Tokens s, s)
 grabLine ts =
   -- remove the newline from the stream
-  case take1_ ts of
+  case take1_ ts' of
       Nothing    -> (line, ts')
       Just (_,v) -> (line, v  )
   where
     -- grab the line content
     (line, ts') = takeWhile_ f ts
-    f x =
-      case fromEnum x of
-          10 -> False
-          _  -> True
+    f x = fromEnum x /= 10
 
 -- | Pretty-print a stack of source positions.
 --
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -57,7 +57,7 @@ import Text.Megaparsec.Pos
 import Text.Megaparsec.Stream 
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Set           as E
-import Debug.Trace
+
 #if !MIN_VERSION_base(4,8,0)
 import Control.Applicative
 #endif
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -15,6 +15,7 @@
 -- You probably do not want to import this module directly because
 -- "Text.Megaparsec" re-exports it anyway.
 
+
 {-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DeriveDataTypeable  #-}
 {-# LANGUAGE DeriveFunctor       #-}
@@ -32,6 +33,7 @@ module Text.Megaparsec.Error
   , ShowErrorComponent (..)
   , parseErrorPretty
   , parseErrorPrettyWithLine
+  , parseErrorPrettyWithStreamContext
   , sourcePosStackPretty
   , parseErrorTextPretty )
 where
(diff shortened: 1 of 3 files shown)


recursion-ninja:
This still needs to be tested with a stack of source positions that has a depth greater than one.

We might also want to consider when a parser error contains multiple unexpected tokens, underlining each unexpected token that is on the same line as the first unexpected token.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
 parseErrorPretty e =
   sourcePosStackPretty (errorPos e) <> ":\n" <> parseErrorTextPretty e
 
+
+-- | Pretty-print a 'ParseError' and display the line on which the parse error
+-- occurred. The rendered 'String' always ends with a newline.
+--
+-- Assumes that the stream tokens can be coerced to an @Int@ and that the value
```
mrkkrp: I think "converted" is a better word to use here, as "coerced" implies that we use actual coercion between types that is supported by GHC (`coerce` for type-safe coercion and `unsafeCoerce` for unsafe coercion), which is not the case.

You could also mention `fromEnum`, so it's clear how the "conversion" is performed.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+streamIter f p ts = fst $ go (x-1) (f ts)
+  where
+    x = unPos p
+    go 0 val    = val
+    go n (_,xs) = let val = f xs
+                  in go (n-1) val
+
+-- | Takes a line from a stream.
+--
+-- Tokens which are numericaly coercable to the integer value @10@ are considered
```
mrkkrp: Again here "coercable" -> "convertable".

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+grabLine ts =
+  -- remove the newline from the stream
+  case take1_ ts' of
+      Nothing    -> (line, ts')
+      Just (_,v) -> (line, v  )
+  where
+    -- grab the line content
+    (line, ts') = takeWhile_ isNotNewLine ts
+
+    -- Maybe instead we want to corece to an @Int@, then convert to a @Char@,
```
mrkkrp: No, that'll be most certainly slower, and `defaultAdvance1` is very sensitive to slow-downs because it's called on every token. Comparison with plain number 10 is OK.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+    cs =
+      case ts of
+        [] -> "<empty line>"
+        xs -> chr . fromEnum <$> xs
+    ts = chunkToTokens (Proxy :: Proxy s) lineVal
+    lineVal = streamIter grabLine linePos s
+    linePos = sourceLine p
+    columnNum = unPos $ sourceColumn p
+
+-- | Applies a "taking" function @n@ number of times to a 'Stream' and returns
```
mrkkrp: You need to escape the double quote character even though this Haddock is not public, it's a good idea to write correct Haddock markup anyway:

> Applies a \"taking\" ...

And there is a typo on the second like in `"tking"`.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+    lineNumberStr = show $ unPos linePos
+    padLength = length lineNumberStr + 2
+    padding = replicate padLength ' '
+
+    cs =
+      case ts of
+        [] -> "<empty line>"
+        xs -> chr . fromEnum <$> xs
+    ts = chunkToTokens (Proxy :: Proxy s) lineVal
+    lineVal = streamIter grabLine linePos s
```
mrkkrp: I see `streamIter` and `grabLine` are only used once here, for dropping `linePos` lines from the beginning of the stream. I would like you to merge the both functions into a single function that just clearly does what we need here: dropping N lines and returning the line we arrive at after that. This will be more clear to future readers of the code.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+  :: ( Enum (Token s)
+     , Stream s)
+  => (s -> (Tokens s, s)) -- ^ A function which takes part of the stream
+  -> Pos                  -- ^ Number of times to apply the function.
+  -> s                    -- ^ Stream to be taken from
+  -> Tokens s             -- ^ The the last segment taken from the stream
+streamIter f p ts = fst $ go (x-1) (f ts)
+  where
+    x = unPos p
+    go 0 val    = val
```
mrkkrp: Don't forget to add strictness annotation to the both arguments.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+    paddingLine   = padding <> "|"
+    contextLine   = " " <> lineNumberStr <> " | " <> cs
+    pointingLine  = padding <> "|" <> replicate columnNum ' ' <> "^"
+
+    lineNumberStr = show $ unPos linePos
+    padLength = length lineNumberStr + 2
+    padding = replicate padLength ' '
+
+    cs =
+      case ts of
```
mrkkrp: Two things I'm not sure about:

1. Is printing `"<empty line>"` better than just printing empty line?
2. How to handle the cases when we point to a `'\n'` character? Probably we should place the caret `^` to the end of line the `'\n'` character ends? This is a corner case we should test.

Text/Megaparsec/Error.hs
```diff
@@ -248,6 +252,98 @@ parseErrorPretty
...
+  => (s -> (Tokens s, s)) -- ^ A function which takes part of the stream
+  -> Pos                  -- ^ Number of times to apply the function.
+  -> s                    -- ^ Stream to be taken from
+  -> Tokens s             -- ^ The the last segment taken from the stream
+streamIter f p ts = fst $ go (x-1) (f ts)
+  where
+    x = unPos p
+    go 0 val    = val
+    go n (_,xs) = let val = f xs
+                  in go (n-1) val
```
mrkkrp: I'm not entirely sure why this can't be written just as `go (n-1) (f xs)`, but this is not really important.

mrkkrp:
Don't mind Travis, it has a weird spread of build times ranging from 20 minutes or less to 50 minutes where the process gets killed.

mrkkrp:
> We might also want to consider when a parser error contains multiple unexpected tokens

This is no longer possible in version 6. I'll write a blog post about that, but in short we always have one unexpected token at the moment of parse error generation (makess sense, you consider "unexpected" what you run into, and you can't run into two different things at the same position at once) and several unexpected tokens only happen because of merging of different parse errors from different branches of parsing. I had to change the merging algorithm to make parse errors generated by the new `tokens` function merge more nicely and so some normalization were required. That normalization however prevents us from having `ParseError` as lawful monoid (which we would like very much) unless we switch from set of unexpected tokens to `Maybe`.

recursion-ninja:
Thanks for all the feedback. I'll address all the issues you made note of.

I ran my branch on a bio-informatics input file that I placed a few parse errors in and the rendering was less than ideal. The input file contains segments of a genome and has lines within the file that are 10,000 characters long. Rendering out the whole 10,000 character line isn't terribly informative so I thought that we could pass in a "token range" parameter that says, render at most this many tokens around the line of the parse error. A range of 80 or 120 should be appropriate for most files, but the user can specify the token range manually if they want. 

I think this can generalize to work on a stream of arbitrary tokens, not just with "numerically-Char-like" tokens. Let me know what you think. I'll have time to try implementing it in about 12 hours.

mrkkrp:
Yeah "token range" is a good idea.

mrkkrp:
Any progress on this?

recursion-ninja:
I'm about halfway through adding the "token range." I added a "total-tokens-consumed" field to the parse error to make things easier in the error message post-processing with the stream in context.

mrkkrp:
>  I added a "total-tokens-consumed" field to the parse error to make things easier

We can't go that way, IMO. The field is essentially useless because the total number of processed tokens is an accumulated value from entire parser run, while this use case demands something tied to an individual file. The difference is apparent when we have include files or something like that.

mrkkrp:
I'm planning to push towards release of version 6 next week. There are also a couple of libraries I want to upgrade to Megaparsec 6 soon and a new project that requires it. Would it be possible for you to push what you have so far and I could pick up your ideas and finish the PR? That's in case you don't have time to finish yourself in a couple of days.

Sorry to hurry you :-|

recursion-ninja:
No worries about "hurrying" me. I think I got the core contextualized rendering complete. I haven't had sufficient time to test all the corner cases yet nor address some of the code review comment you made earlier. I pushed what I have so far.

Hopefully I can add a test suite and remove any defects over the next week, along with improve the documentation and address all code review comments had/have. Feel free to move forward with what I have so far.

recursion-ninja:
> We can't go that way, IMO. The field is essentially useless because the total number of processed tokens is an accumulated value from entire parser run, while this use case demands something tied to an individual file. The difference is apparent when we have include files or something like that.

Do you have an example somewhere demonstrating this kind of parsing case? I'm not sure I follow, and would like to poke around with an example if possible so I can better understand the issue.

mrkkrp:
Imagine you parse file A and you have processed 10 tokens from there. Next you run into an include pragma and so you set new input (file B) and use [`pushPosition`](https://hackage.haskell.org/package/megaparsec-5.3.1/docs/Text-Megaparsec-Prim.html#v:pushPosition) to set position at the beginning of the file. The counter for the total number of processed tokens isn't reset however, so if at this point a parse error occurs in file B, you'll have greater number of processed tokens than you expect (greater by 10).

mrkkrp:
I'll start working on finishing this later today (notifying you so we don't duplicate effort). I'll add tests too.

mrkkrp:
To be continued in #236.

mrkkrp:
@recursion-ninja, How would you like to be added to the `AUTHORS.md` file?

recursion-ninja:
```Alex Washburn (@recursion-ninja)```

PR #236, opened 2017-07-23
Implement rendering of parse error context

### LINKED ISSUE #205: Implement optional displaying of releavant region of input in error messages

mrkkrp:
Resolves #205, continuation of #233.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -24,6 +24,7 @@ Names below are sorted alphabetically.
 ## Contributors
 
 * Albert Netymk
+* Alex Washburn (@recursion-ninja)
 * Antoine Latter
 * Artyom (@neongreen)
 * Auke Booij
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,17 @@
   we had something like that in the `hspec-megaparsec` package, but it does
   not hurt to ship it with the library.
 
+* Added `parseErrorPretty'` allowing to display offending line in parse
+  errors.
+
+* Added `LineToken` type class for tokens that support operations necessary
+  for selecting and displaying relevant line of input (used in
+  `parseErrorPretty'`).
+
+* Added `parseTest'` function that is just like `parseTest`, but also prints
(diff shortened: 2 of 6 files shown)


mrkkrp:
@recursion-ninja, This is a simpler version without the setting that controls max line width. My logic was that lines as long as several thousand characters are not common, and we'll be able to add more functionality in backward-compatible way in 6.x.x series of the library. Let's start with basic line rendering for now.

recursion-ninja:
I saw that. I know that the bio-informatics program I'm working on often accepts files with *very* long lines, but I'm not sure how common that is in general.

I wish I had more time to get the more general rendering working. I'll keep poking around with the more general approach. Hopefully, I can do that without the problematic "total tokens consumed" field.

recursion-ninja:
It just dawned on me, that I could just use my version of `parseErrorPrettyWithContex` defined outside of the `megaparsec` library on the 1000+ line length bio-informatics files. 

Do you think there is value in trying to include this kind of error rendering in a later release, or should I just have a custom renderer for my program?

mrkkrp:
I'm honestly not sure how common this use case is. I think it's best for you to use your custom solution. See how things are done in the released version of library, perhaps you could adjust `parseErrorPrettyWithContext` to use `LineToken` type class, etc. Then we can try to include it in the library.

PR #242, opened 2017-08-08
Clarify the difference between `pure` and `empty`

Profpatsch:
The first consumes nothing and succeeds, the second fails.
This is important for e.g. the `space` lexer arguments,
which is also added to the documentation.

### DIFF
--- a/Text/Megaparsec.hs
+++ b/Text/Megaparsec.hs
@@ -311,6 +311,7 @@ pMap f p = ParsecT $ \s cok cerr eok eerr ->
   unParser p s (cok . f) cerr (eok . f) eerr
 {-# INLINE pMap #-}
 
+-- | 'pure' will return a parser that __succeeds__ but consumes no input.
 instance Stream s => A.Applicative (ParsecT e s m) where
   pure     = pPure
   (<*>)    = pAp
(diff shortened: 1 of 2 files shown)


Text/Megaparsec.hs
```diff
@@ -329,10 +330,12 @@ pAp m k = ParsecT $ \s cok cerr eok eerr ->
   in unParser m s mcok cerr meok eerr
 {-# INLINE pAp #-}
 
+-- | 'A.empty' will return a parser that __fails__ and consumes no input.
 instance (Ord e, Stream s) => A.Alternative (ParsecT e s m) where
   empty  = mzero
   (<|>)  = mplus
 
+-- | See comment for 'pure'.
```
mrkkrp: I think it's better to be a bit repetitive here (if we start to give this sort of clarification):

> 'return' returns a parser that __succeeds__ without consuming input.

Text/Megaparsec.hs
```diff
@@ -399,6 +402,7 @@ instance (Stream s, MonadError e' m) => MonadError e' (ParsecT e s m) where
     runParsecT p s `catchError` \e ->
       runParsecT (h e) s
 
+-- | See comment for 'A.empty'.
```
mrkkrp: Similarly here let's be explicit about `mzero`.

Text/Megaparsec/Char/Lexer.hs
```diff
@@ -91,6 +91,9 @@ import qualified Text.Megaparsec.Char as C
 -- 'skipBlockComment' or 'skipBlockCommentNested' if you don't need anything
 -- special.
 --
+-- If you don't have a parser for a kind of comment,
```
mrkkrp: Maybe it's worth changing `use 'empty' as placeholder` to `pass 'empty'` or `use 'empty'`. It's nitpicking, but `empty` in this case is not a placeholder, it's just a parser that instantly fails and lets other things run instead.

mrkkrp:
I went ahead and applied as c8ed8d0112da3e00d08503ca18cd990e8bfbf4dc to save time. Thanks again!

PR #256, opened 2017-11-16
Improve documentation of lexeme conventions

ntc2:
I DID NOT TEST GENERATING THE HADDOCKS TO CONFIRM THIS RENDERS CORRECTLY.

Move the discussion of lexeme assumptions from the bottom of the already very long documentation for `space` to the top-level module documentation, and bold the key convention that lexeme parsers assume no space before and consume all space after the lexeme they parse.

### DIFF
--- a/Text/Megaparsec/Char/Lexer.hs
+++ b/Text/Megaparsec/Char/Lexer.hs
@@ -14,6 +14,14 @@
 -- more elegant than others. Especially important theme is parsing of white
 -- space, comments, and indentation.
 --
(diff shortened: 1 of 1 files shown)


mrkkrp:
Thank you, this is a good change. Unfortunately Haddock is not smart enough to understand that bold text can contain a newline in the middle, so I too the liberty of amending your commit to get a proper rendition.

Applied as: 8dc71651cf673af511cdc6543e540e42a148686f.

PR #261, opened 2017-12-15
Export token-independent combinators from Text.Megaparsec

int-index:
There's one inconsistency and one inconvenience when it comes to the module structure of megaparsec.

* The inconsistency is that some combinators are exported from `Text.Megaparsec` (i.e. `match`, `takeRest`), while some are exported from `Text.Megaparsec.Char` (i.e. `satisfy`), although there is nothing Char-specific in `satisfy`.

* The inconvenience is that almost always there are two imports required to use `megaparsec`: I need to import both `Text.Megaparsec` for core definitions and `Text.Megaparsec.Char` for Char-specific definitions.

I figured out a simple way to address both concerns at once, and have a principled way to structure the modules. The principle is this:

* all combinators that do not assume a specific token type go into `Text.Megaparsec`
* combinators that assume a specific token type (i.e. `Token s ~ Char`) go into a corresponding module

Because of this, I moved `char`, `oneOf`, `noneOf`, `satisfy`, etc, to `Text.Megaparsec`. This eliminated the awkward fact that `Text.Megaparsec.Byte` was dependent on `Text.Megaparsec.Char`, while those modules are actually for completely different input streams and, in principle, are independent.

Finally, I re-exported `Text.Megaparsec` from `Text.Megaparsec.Char` and `Text.Megaparsec.Byte`, because this way only one import is needed. I cannot imagine a case where a user would want to use `Text.Megaparsec.Char` without core definitions from `Text.Megaparsec`.

The breakage is minimal (there might be a few name conflicts, because `Text.Megaparsec` exports more combinators now), but this should not be a problem.

This makes the module structure more principled and ergonomic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Megaparsec 6.4.0
+
+* Exported combinators independent of token type from `Text.Megaparsec`,
+  re-exported `Text.Megaparsec` from `Text.Megaparsec.Char` and
+  `Text.Megaparsec.Byte`.
+
 ## Megaparsec 6.3.0
 
 * Added an `IsString` instance for `ParsecT`. Now it is possible to
--- a/Text/Megaparsec.hs
+++ b/Text/Megaparsec.hs
@@ -101,6 +101,14 @@ module Text.Megaparsec
   , region
   , takeRest
   , atEnd
+  , char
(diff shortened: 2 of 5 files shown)


mrkkrp:
I commented on this in a private chat.

Issue #289, opened 2018-04-20
sepBy/manyTill missing


flip111:
https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec.html#v:parse

https://markkarpov.com/megaparsec/parsing-simple-imperative-language.html

On those two pages there is a mention of `sepBy` and `sepBy1`. However i can't find these functions in megaparsecc 6.5.0 documentation.

https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec-Char-Lexer.html#v:charLiteral

https://hackage.haskell.org/package/megaparsec-6.5.0/docs/doc-index-M.html

manyTill is missing too

mrkkrp:
They all are here: https://hackage.haskell.org/package/parser-combinators. Since version 6.

mrkkrp:
The combinators are re-exported for you from `Text.Megaparsec`.

flip111:
Is it possible to link to that documentation in more places ? It's difficult to find i think

mrkkrp:
Well it's almost at the top already: https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec.html, section about re-exports.

flip111:
Maybe make a list of functions that are re-exported and put them in the documentation so people can search by function name.

mrkkrp:
Then I’ll need to remember to keep the export list in sync with the `parser-combinators` package, which is error-prone.

recursion-ninja:
@flip111 Note that if you look at the haddock documentation for `[Text.Megaparsec`](https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec.html) you can select the [`index`](https://hackage.haskell.org/package/megaparsec-6.5.0/docs/doc-index.html) link in the upper right hand corner. This takes you to a page where you can search _every_ exported function of the package from _every_ module combined either by letter or view the all functions on a single page by selecting [`All`](https://hackage.haskell.org/package/megaparsec-6.5.0/docs/doc-index-All.html). 

It is possible to discover the exported functions by using the standard Haddock documentation.

mrkkrp:
Closing because I don’t see what action should be taken here. We already have a section in the docs about the re-export.

PR #290, opened 2018-04-20
Update Megaparsec.hs

flip111:
https://github.com/mrkkrp/megaparsec/issues/289


mrkkrp:
See comment in #289.

flip111:
This should add a section "Combinators" to the contents (block on the top right) of https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec.html it's useful for when people are looking for the combinators. It's another way to get to the same information, which is good because it makes things easier to find. (By more ways to find something = more easier to find it)

PR #297, opened 2018-05-17
Fix `charLiteral` for long character escape sequences

tvh:
I could trigger it with the following: `char '"' *> manyTill L.charLiteral (char '"')` and the imput `"\154972\&1"`


mrkkrp:
Note that 9 appears to be quite enough here. The `\&` part is just a separator between the char code before it and normal next after it so input after `\&` is not important. Thus we should grab 9 characters at most.

Here is a test case that I added (passes with 9):

```haskell
    context "when given a long escape sequence" $
      it "parses it correctly" $
        property $ \s' -> do
          let s = "\\154972\\&" ++ s'
          prs charLiteral s `shouldParse` '\154972'
          prs' charLiteral s `succeedsLeaving` s'
```

tvh:
For this particular testcase 9 is enough, but 10 is the maximum:
"\1114111\&1"

tvh:
10 is also the exact length you get to when adding this corner case to the previous limit

mrkkrp:
Great, I'll update the test then.

PR #300, opened 2018-05-29
Fixed typo (staring -> 'String')

sshine:
Big fan of Megaparsec. :)


mrkkrp:
The failure has nothing do to with the PR, obviously. Looks like another Unicode thing. I took a note of it and will investigate when I have time. Restarted the job.

PR #313, opened 2018-08-17
Use offset in parse errors

mrkkrp:
The change allows to speed-up parsers in most cases and re-use the fact that we're traversing input stream on pretty-printing `ParseError`s anyway.

TODO:

- [x] Adjust the rest of old tests, add new tests.


mrkkrp:
So far benchmarks show very promising improvements and overall this feels like a right thing to do.

PR #317, opened 2018-08-28
UndecidableInstances needed in one module for GHC 8.6.

galenhuntington:
See https://ghc.haskell.org/trac/ghc/wiki/Migration/8.6#UndecidableInstancesispickier.

With this change, I am able to build on GHC 8.6.1-beta1.  Without it, the four uses of `Token s` in instance heads give errors.


mrkkrp:
Thanks!

PR #327, opened 2018-09-23
More descriptive error when elabel or ulabel is used incorrectly

### LINKED ISSUE #325: Add more descriptive errors for elabel and ulabel
When writing a test earlier today, I used `elabel` with `""` (silly I know, but it was a WIP 😅), but it gave me a non-descriptive error which basically boiled down to:

```
uncaught exception: ErrorCall
NonEmpty.fromList: empty list
```

A quick search through to the codebase shows that both `ulabel` and `elabel` directly use `NE.fromList` without checking if the list is indeed empty.
Best to add empty list check before constructing those labels?

I could add the change myself, but what is the behaviour you prefer? Raising custom exception on empty label?

luc-tielen:
Fixes #325.


mrkkrp:
Thanks!

PR #328, opened 2018-10-03
Add decimalGroup lexer to allow underscores

tomphp:
This PR adds a new `digits'` function to allow the parsing of integers with underscores in them.

Some examples of where this style of number is used:
- New GHC `>8.6.1` (See https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0009-numeric-underscores.rst#validity-examples)
- Languages such as Ruby
- TOML (See tomland which uses megaparsec https://github.com/kowainik/tomland/pull/97)

This only does integers so far, and I'll need some help making this performant but I'm opening the PR now to see if there's any appetite for it.

### DIFF
--- a/Text/Megaparsec/Char/Lexer.hs
+++ b/Text/Megaparsec/Char/Lexer.hs
@@ -54,6 +54,7 @@ module Text.Megaparsec.Char.Lexer
   , charLiteral
     -- * Numbers
   , decimal
+  , groupedDecimal
   , binary
   , octal
   , hexadecimal
@@ -62,7 +63,7 @@ module Text.Megaparsec.Char.Lexer
   , signed )
 where
 
-import Control.Applicative
+import Control.Applicative hiding (many)
 import Control.Monad (void)
 import Data.List (foldl')
 import Data.List.NonEmpty (NonEmpty (..))
@@ -347,6 +348,27 @@ decimal_ = mkNum <$> takeWhile1P (Just "digit") Char.isDigit
     mkNum    = foldl' step 0 . chunkToTokens (Proxy :: Proxy s)
     step a c = a * 10 + fromIntegral (Char.digitToInt c)
 {-# INLINE decimal_ #-}
+
+-- | Same as 'decimal' but allows underscores as separators in the string.
+
+groupedDecimal :: (MonadParsec e s m, Token s ~ Char, Integral a) => m a
+groupedDecimal = groupedDecimal_ <?> "integer"
+{-# INLINEABLE groupedDecimal #-}
+
+-- | A non-public helper to parse decimal integers with underscores.
+
+groupedDecimal_
+  :: forall e s m a. (MonadParsec e s m, Token s ~ Char, Integral a)
+  => m a
+groupedDecimal_ = mkNum . fst <$> match (do void $ digits
+                                            void $ many $ try _digits
+                                            notFollowedBy (C.char '_'))
+  where
+    digits   = takeWhile1P (Just "integer") Char.isDigit
+    _digits  = takeWhile1P (Just "underscore") (== '_') >> digits
+    mkNum    = foldl' step 0 . chunkToTokens (Proxy :: Proxy s)
+    step a c = if c == '_' then a else a * 10 + fromIntegral (Char.digitToInt c)
+{-# INLINE groupedDecimal_ #-}
 
 -- | Parse an integer in binary representation. Binary number is expected to
 -- be a non-empty sequence of zeroes “0” and ones “1”.
--- a/bench/memory/Main.hs
+++ b/bench/memory/Main.hs
@@ -47,6 +47,7 @@ main = mainWith $ do
   bparser "takeWhileP" manyAs (const $ takeWhileP Nothing (== 'a'))
   bparser "takeWhile1P" manyAs (const $ takeWhile1P Nothing (== 'a'))
   bparser "decimal" mkInt (const (L.decimal :: Parser Integer))
+  bparser "groupedDecimal" mkInt (const (L.groupedDecimal :: Parser Integer))
   bparser "octal" mkInt (const (L.octal :: Parser Integer))
   bparser "hexadecimal" mkInt (const (L.hexadecimal :: Parser Integer))
   bparser "scientific" mkInt (const L.scientific)
--- a/bench/speed/Main.hs
+++ b/bench/speed/Main.hs
@@ -45,6 +45,7 @@ main = defaultMain
   , bparser "takeWhileP" manyAs (const $ takeWhileP Nothing (== 'a'))
   , bparser "takeWhile1P" manyAs (const $ takeWhile1P Nothing (== 'a'))
   , bparser "decimal" mkInt (const (L.decimal :: Parser Integer))
+  , bparser "groupedDecimal" mkInt (const (L.groupedDecimal :: Parser Integer))
   , bparser "octal" mkInt (const (L.octal :: Parser Integer))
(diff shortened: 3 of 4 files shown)


TikhonJelvis:
The general idea makes sense to me, but the name `decimal'` is not very descriptive. I don't have a great alternative in mind, but I could see something like `groupedDecimal`.

tomphp:
Good to hear @TikhonJelvis - I've updated the name to `groupedDecimal` now.

Before I continue to try and do the other number formats, do you have any suggestions on implementation (I'm not hugely familiar with `megaparsec` and would like to do it in the most idiomatic way possible).

mrkkrp:
@tomphp So have you given up on this so quickly? If so, I guess we'd need to get this in proper shape and merge ourselves, or close the PR.

tomphp:
Sorry @mrkkrp - I've not forgotten this, I just haven't had the opportunity to look at it. Will do asap.

tomphp:
Current benchmarks with no grouping (i.e. no `_` characters) - how does this look?

```
benchmarking decimal/500
time                 179.2 ns   (178.3 ns .. 180.2 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 179.3 ns   (178.5 ns .. 180.3 ns)
std dev              2.909 ns   (2.141 ns .. 4.534 ns)
variance introduced by outliers: 19% (moderately inflated)

benchmarking decimal/1000
time                 290.2 ns   (288.9 ns .. 291.6 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 290.0 ns   (289.1 ns .. 291.4 ns)
std dev              3.806 ns   (3.068 ns .. 4.664 ns)
variance introduced by outliers: 13% (moderately inflated)

benchmarking decimal/2000
time                 520.0 ns   (516.8 ns .. 524.1 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 520.7 ns   (518.3 ns .. 523.3 ns)
std dev              8.658 ns   (7.238 ns .. 10.82 ns)
variance introduced by outliers: 19% (moderately inflated)

benchmarking decimal/4000
time                 1.179 μs   (1.173 μs .. 1.185 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.178 μs   (1.174 μs .. 1.185 μs)
std dev              18.67 ns   (15.34 ns .. 23.67 ns)
variance introduced by outliers: 16% (moderately inflated)

benchmarking groupedDecimal/500
time                 262.7 ns   (261.6 ns .. 263.7 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 262.2 ns   (261.3 ns .. 263.5 ns)
std dev              3.746 ns   (2.901 ns .. 5.375 ns)
variance introduced by outliers: 15% (moderately inflated)

benchmarking groupedDecimal/1000
time                 374.8 ns   (373.0 ns .. 377.0 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 375.2 ns   (373.5 ns .. 377.7 ns)
std dev              6.800 ns   (4.819 ns .. 10.54 ns)
variance introduced by outliers: 21% (moderately inflated)

benchmarking groupedDecimal/2000
time                 630.6 ns   (628.0 ns .. 633.2 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 625.2 ns   (622.0 ns .. 628.4 ns)
std dev              11.08 ns   (9.513 ns .. 12.94 ns)
variance introduced by outliers: 20% (moderately inflated)

benchmarking groupedDecimal/4000
time                 1.315 μs   (1.309 μs .. 1.322 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.310 μs   (1.303 μs .. 1.317 μs)
std dev              21.86 ns   (16.85 ns .. 28.76 ns)
variance introduced by outliers: 17% (moderately inflated)
```

mrkkrp:
Doesn't look too bad, I'll review when I have time then merge and release a new version.

tomphp:
I also want to do the hexedecimal, float and scientific - but happy to do them in separate PRs if you want to get this one merged asap.

mrkkrp:
Hmm, I'm not sure we should add many such functions because the use case in rather marginal already.

tomphp:
Ok @mrkkrp - I'm happy to leave it just as this, I just thought it was worth doing as they are all supported by GHC now.

TikhonJelvis:
Would it make sense to extract this into a combinator like `underscoreGrouping`? I'm imagining `underscoreGrouping decimal`, `underscoreGrouping binary`... etc.

This will help us avoid needing a bunch of new functions for this, but I'm not sure if it would be intuitive from an API-design point of view or if it could be implemented efficiently.

mrkkrp:
@tomphp Can you try to refactor the code to add such an operator?

tomphp:
Yes, I'm happy to do that @mrkkrp.

Quick note: I think it will work fine for `decimal` and `binary`, but wouldn't work so well float `float`, `hexidecimal` or `scientific`

tomphp:
My mistake @mrkkrp, I had assumed it parsed the `0x` prefix - it should be fine ;-)

tomphp:
I've just started looking at this not and I don't really know how to do it. The `decimal` and `binary` functions each do two things - they parse the string and then `fmap` their `mkNum` functions over the result. In order to create the `underscoreGrouping` function, I need to combine the existing parser with the new underscoring bit, but keep the `mkNum` function and I can't see how to do it?

Is there an example where something like this has been done before?

TikhonJelvis:
I'm not sure if it's possible either :). I'll take a deeper look at it to see if I can figure something out.

If I can't, what's the default plan? Supporting the functionality for decimal but not for the other options? Would it make sense to have combinators that take a base as an argument?

TikhonJelvis:
Okay, I've looked at this in more depth and couldn't figure out how to write a combinator for this. I suspect it wouldn't be possible without more capabilities in the `Stream` class (ie some way to filter a stream before running a parser and recover the remainder without the filter).

tomphp:
Thanks for looking into this @TikhonJelvis.

How should I proceed with this now?

mrkkrp:
I mainly don't like adding three very similar combinators because of code duplication. And if we can't have a combinator to modify the existing ones, then it's too bad.

I'll take a look myself and try to refactor this a bit this week. I think the code can be 1) improved, there are bits in `_groupedDecimal` that I don't like now that I'm looking at it more carefully, `match` certainly shouldn't be there 2) the helper that is currently called `groupedDecimal_` can be parametrized over what we have right now as `Char.isDigit` and perhaps then we could have all the different versions of this.

mrkkrp:
I decided to close this one without adding anything. The reason is that if we're going to add this feature, then we'd need to support everything described in [this proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0009-numeric-underscores.rst#validity-examples) properly, which at present is going to introduce a bit of duplication and will double API for numeric parsers (we'd also need to support underscores in floating point numbers).

As I see it, we'll also need to add everything 2 times, for character and binary streams. The situation is already unfortunate (still the best I could do), let's not make it worse. The change would also increase maintaining burden.

I see value in this feature, but it's just not worth adding it to the main library. Projects that need to parse such numerals will need to write their own helpers. Or there may be a helper library separate from Megaparsec.

Sorry to disappoint and for the delays. Hope this is understandable.

tomphp:
Totally understandable @mrkkrp - thanks for taking the time to consider it.

PR #331, opened 2018-10-31
Fix build with mtl < v2.2.2

frasertweedale:
Commit 56c89cba24903a966678c3508deeff8a0af0b68e removed a CPP
conditional.  mtl Control.Monad.Identity only exports
Control.Monad.Trans.Identity as of v2.2.2.  Therefore IdentityT is
not in scope with mtl < v2.2.2. and the build fails.

Restore the CPP block to fix building with mtl < v2.2.2.

### DIFF
--- a/Text/Megaparsec/Class.hs
+++ b/Text/Megaparsec/Class.hs
@@ -14,6 +14,7 @@
 --
 -- @since 6.5.0
 
+{-# LANGUAGE CPP                    #-}
 {-# LANGUAGE FlexibleInstances      #-}
 {-# LANGUAGE FunctionalDependencies #-}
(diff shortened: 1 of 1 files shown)


mrkkrp:
Do you want me to cut a new release after this is merged?

mrkkrp:
Thanks! I’ll cut a release once I’m home.

hvr:
@frasertweedale so this means that 7.0.2 needs its `mtl` lower bound tightened?

frasertweedale:
@hvr yes, to >=2.2.2.  Or suffer the build failures...

mrkkrp:
I tightened the bounds for 7.0.2.

Issue #255, opened 2017-11-02
`instance Ord a => Stream [a]`


tdammers:
This would be terribly useful for situations where you have a separate lexer, so that your parser becomes something like `[Lexeme] -> ParseTree`; but also for situations where your input is not a stream of bytes or characters, but, say, a stream of strings (e.g. for parsing CLI options). Most of the `String` instance, I believe could be reused, the only thing char-specific about it is the handling of newlines.

mrkkrp:
I'm not sure how `defaultAdvance1` is going to work with such an abstract instance. If all we know about `a` is that it is an instance of `Ord`, I think we can't implement correct position-advancing methods of `Stream`. You're welcome to try though, I'm open to reviewing a PR that adds such an instance.

As for the use cases you mention, I would favor `newtype` wrappers for that.

tdammers:
Well, the idea would obviously be that the position-advancing would simply increment the source column, and keep the source line at 1, since there is no notion of "newlines" for arbitrary tokens. This conflicts with the String instance though, so I agree that a newtype wrapper would be in order. Might still be useful to have a generic implementation of such a wrapper in the library though, after all this doesn't seem like an outrageous use case to me.

mrkkrp:
If we just increment source column by 1 per token, wouldn't it be confusing? Because even with the examples you mention (a list of lexemes and a list of strings), a token (lexeme/string) is certainly capable of being "wider" than 1 column.

I think the problem here is that the library can parse things that are not tied to the notions of line and column, but the `SourcePos` type is currently defined so that this "system of coordinates" is imposed on everything we accept as input. Not sure what to do about that.

tdammers:
So far, my mental model is such that the newtype wrapper would imply the semantics of source positions, and there are two solutions that make sense to me:

1. Every lexeme is considered exactly one unit wide, the lexeme stream is considered 1-dimensional, and "column" are considered in "lexeme space" (that is, column 24 refers to the 24th lexeme). This is easy, we'd just increment the column count by 1 for each lexeme. This is most useful for things like argument parsers, where we don't have 2-dimensional input anyway, and numbering lexemes matches the problem domain's semantics ("5th argument" makes sense for CLI commands, but line numbers really don't).
[...]

PR #335, opened 2018-12-08
[WIP] Add Stream instance to any list

Infinisil:
This is an already working implementation for being able to use arbitrary lists as token streams. The only constraint elements need is to be an instance of `SrcPosToken`, which provides functions for getting newlines/tabs and pretty printing. Default implementations for all functions is provided, so it's just a matter of declaring the instance:

```haskell
instance SrcPosToken Int
```

Then you can have parsers of `Int` lists:
```haskell
lengthPrefixed :: Parsec Data.Void.Void [Int] [Int]
lengthPrefixed = do
  len <- satisfy (>=0)
  replicateM len anySingle

result = parse (many lengthEncoded) "" [2, 1, 10, 3, 1, 2, 3]
-- Right [[1, 10], [1, 2, 3]]
```

Unfortunately you get an orphan instance warning if you do this though, which I don't know how to prevent in a convenient way:
```
    Orphan instance: instance SrcPosToken Int
    To avoid this
        move the instance declaration to the module of the class or of the type, or
        wrap the type with a newtype and declare the instance on the new type.
```

Closes https://github.com/mrkkrp/megaparsec/issues/255

This is not ready because I'm not sure if I made some bad decisions somewhere, haven't tested it or benchmarked it (it should only affect error construction times though), and maybe some other code could be refactored with this change.

### DIFF
--- a/Text/Megaparsec/Stream.hs
+++ b/Text/Megaparsec/Stream.hs
@@ -15,6 +15,7 @@
 -- @since 6.0.0
 
 {-# LANGUAGE CPP                 #-}
+{-# LANGUAGE DefaultSignatures   #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE LambdaCase          #-}
@@ -25,27 +26,25 @@
 {-# LANGUAGE TypeFamilies        #-}
 
 module Text.Megaparsec.Stream
-  ( Stream (..) )
+  ( Stream (..), SrcPosToken )
 where
 
-import Data.Char (chr)
-import Data.Foldable (foldl')
-import Data.List.NonEmpty (NonEmpty (..))
-import Data.Maybe (fromMaybe)
-import Data.Proxy
-import Data.Word (Word8)
-import Text.Megaparsec.Pos
-import Text.Megaparsec.State
-import qualified Data.ByteString            as B
-import qualified Data.ByteString.Char8      as B8
-import qualified Data.ByteString.Lazy       as BL
-import qualified Data.ByteString.Lazy.Char8 as BL8
-import qualified Data.List.NonEmpty         as NE
-import qualified Data.Text                  as T
-import qualified Data.Text.Lazy             as TL
+import qualified Data.ByteString       as B
+import qualified Data.ByteString.Lazy  as BL
+import           Data.Char             (chr)
+import           Data.Foldable         (foldl')
+import           Data.List.NonEmpty    (NonEmpty (..))
+import qualified Data.List.NonEmpty    as NE
+import           Data.Maybe            (fromMaybe)
+import           Data.Proxy
+import qualified Data.Text             as T
+import qualified Data.Text.Lazy        as TL
+import           Data.Word             (Word8)
+import           Text.Megaparsec.Pos
+import           Text.Megaparsec.State
 
 #if !MIN_VERSION_base(4,11,0)
-import Data.Semigroup
+import           Data.Semigroup
 #endif
 
 -- | Type class for inputs that can be consumed by the library.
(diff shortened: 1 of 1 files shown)


Text/Megaparsec/Stream.hs
```diff
@@ -244,12 +243,12 @@ instance Stream BL.ByteString where
...
   takeWhile_ = BL.span
-  showTokens Proxy = stringPretty . fmap (chr . fromIntegral)
+  showTokens Proxy = prettyTokens . fmap (chr . fromIntegral)
   -- NOTE Do not eta-reduce these (breaks inlining)
   reachOffset o pst =
-    reachOffset' splitAtBL BL.foldl' BL8.unpack (chr . fromIntegral) (10, 9) o pst
+    reachOffset' splitAtBL BL.foldl' ((:[]) . chr . fromIntegral) ((==10), (==9)) o pst
   reachOffsetNoLine o pst =
-    reachOffsetNoLine' splitAtBL BL.foldl' (10, 9) o pst
+    reachOffsetNoLine' splitAtBL BL.foldl' ((==10), (==9)) o pst
```
Infinisil: `Word8` could be made an instance of `SrcPosToken` and then `reachOffset[NoLine]` could avoid having to pass these `(==10)` arguments by using `isLF/isTab`.

mrkkrp:
Thanks. For the record, I've never been super-enthusiastic about #255, I still think it should be handled by `newtype` wrapping. Custom stream of tokens are probably very different anyway from `Char`s because those custom tokens are likely to be more than just characters, i.e. they could be arbitrary lexemes and in that case how do we decide if a lexeme is a tab? Well, it just doesn't make sense. Custom handling is necessary for those cases.

Also honestly I don't like adding another type class for marginal abstraction benefit.

That said, iff this patch doesn't affect performance, I'll take a look later.

Infinisil:
> I still think it should be handled by newtype wrapping.

Yeah I'll have to agree after thinking about it some more. You mean something like this, right? (No ugly warnings here)

```haskell
newtype IntToken = IntToken { unIntToken :: Int } deriving Show

instance SrcPosToken IntToken

lengthPrefixed :: Parsec Data.Void.Void [IntToken] [Int]
lengthPrefixed = do
  len <- unIntToken <$> satisfy ((>=0) . unIntToken)
  replicateM len (unIntToken <$> anySingle)
```

An instance `SrcPosToken Word8` would also be unfitting because these bytes could represent anything and not specifically chars. By letting the user give their own `newtype`, they essentially decide what a tab/line means for it, if there is at all.

> Custom stream of tokens are probably very different anyway from `Char`s because those custom tokens are likely to be more than just characters, i.e. they could be arbitrary lexemes and in that case how do we decide if a lexeme is a tab?

That went into my considerations so I made sure that the default instance acts as if everything is on a single line, meaning `isLF` and co. are just `const False` (this wouldn't be possible if we'd have to provide a newline token for the instance).

> Also honestly I don't like adding another type class for marginal abstraction benefit.

I'm afraid it might be the only way. If you have other ideas please let me know.

mrkkrp:
I meant that streams of custom tokens, even if they are happen to be assembled in a list, need a custom `Stream` instance, so "one-fits-all" instance of `Stream` for lists may be not the best solution.

Infinisil:
@mrkkrp Ohh I see, yeah that sounds better actually. So something like this in Megaparsec

```haskell
newtype ListStream a = ListStream [a]
```

mrkkrp:
Yep.

mrkkrp:
That would be better perhaps, although my point is rather that a user who needs this defines his own `FooStream` wrapper and then writes `Stream` instance for it as he/she wants. I'm not sure how much we can help him/her with this, because every such case is probably rather unique.

PR #336, opened 2018-12-09
[WIP] Mono traversable

### LINKED ISSUE #255: `instance Ord a => Stream [a]`

Infinisil:
I'm just putting this PR here in case I totally stop working on it again tomorrow and forget about it. I just hacked on this for now.

The juicy changes are in `Stream.hs`

Alternative to https://github.com/mrkkrp/megaparsec/pull/335, closes #255 

This completely revamps the `Stream` class, taking almost all methods of it from the [mono-traversable](https://hackage.haskell.org/package/mono-traversable) package instead.

This has the effects:
- Allows easy instances of `Streams`:
  ```haskell
  instance Stream [Int] where
    type Tokens [Int] = [Int]
  ```
  With mono-traversable in addition to the previous `ByteString` and `Text` instances, it also gives us ones for `Vector` and `Seq` for free.
- Gets rid of most methods on `Stream`, delegating it to mono-traversable, which is exactly made for this kind of thing. E.g. `chunkToTokens` is just [`otoList`](https://hackage.haskell.org/package/mono-traversable-1.0.9.0/docs/Data-MonoTraversable.html#v:otoList) and `tokensToChunk` is just [`fromList`](https://hackage.haskell.org/package/mono-traversable-1.0.9.0/docs/Data-Sequences.html#v:fromList).
- Need to use `Element (Tokens s)` instead of `Token s`. That can be circumvented by making an associated type family, but that's a bit hairy (requires `UndecidableInstances`)

I also took this opportunity to get rid of the `Proxy`s, but this is unrelated and I could undo that again. Haven't updated tests or benchmarks or docs.

I should do something about my hlint not messing with other people projects too.


Text/Megaparsec/Stream.hs
```diff
@@ -14,284 +14,102 @@
...
 
-  type Token s :: *
+#if !MIN_VERSION_base(4,11,0)
+import           Data.Semigroup
+#endif
 
-  -- | Type of “chunk” of the stream.
 
+class (IsSequence (Tokens s), Ord (Tokens s), Ord (Element (Tokens s))) => Stream s where
   type Tokens s :: *
```
Infinisil: Here's the main change. Everything else is pretty much just a consequence of this (and my hacking/messing around..)

mrkkrp:
Sorry, we can't make the dep tree even heavier, users are already not happy :-|

mrkkrp:
I'm curious though what happens to benchmarks with this change?

Infinisil:
Eh it's fine, I mainly saw this as an opportunity to learn, and in that sense it was a major success.

The benchmarks should be the same, because mono-traversable uses all the specialized implementations (like Text.span, BS.span, but under a single function). I haven't tested it though.

This approach certainly looks like the right thing to do imo, it's unfortunate that we can't use more dependencies.

PR #348, opened 2019-04-13
Adapt to MonadFail changes in base 4.13.

AndreasPK:


AndreasPK:
Tested it with 8.6.4 and HEAD.

mrkkrp:
Can you please add a bit of background for this change? @hvr could you take a look?

mrkkrp:
AFAIU `fail` will be finally dropped from `Monad` in the next GHC release?

AndreasPK:
> AFAIU `fail` will be finally dropped from `Monad` in the next GHC release?

That is correct. See eg the entry on haskell wiki. https://wiki.haskell.org/MonadFail_Proposal

mrkkrp:
Sounds good, I merged this in.

Issue #315, opened 2018-08-20
Add "commit" combinator


simonmichael:
The 6.0.0 changelog said:

> - Changed signatures of failure and token, they only can signal trivial errors now.
> - Added a new method of MonadParsec type class called fancyFailure for signalling non-trivial failures. Signatures of some functions (failure, token) have been changed accordingly.

What does trivial/non-trivial here mean.. anything to do with backtracking ?

In hledger we sometimes want to throw a final parse error and guarantee that's what the user will see, with no further backtracking. We think fail doesn't do this, and we are [wondering](https://github.com/simonmichael/hledger/issues/853#issuecomment-413513794) if failure/fancyFailure do it. 

hledger benefits hugely from megaparsec, thanks for the excellent package and support.

mrkkrp:
OK, let's see if I can clarify. There are two kinds of errors: trivial and fancy ones. When two parse errors happen exactly at the same location (for example you failed to match `'a'`, hasn't consumed anything, then failed to match `'b'` -- another alternative), they are merged together. Trivial errors only can be merged with trivial errors and fancy errors only can be merged with other fancy errors. Fancy errors overwrite trivial errors if they happen at the same position.

> In hledger we sometimes want to throw a final parse error and guarantee that's what the user will see, with no further backtracking.

There is no way to disable backtracking, but I suspect that most errors are trivial, and if you use `fancyFailure` in this case, it'll overwrite any other trivial errors that happened at the same position, and so the user will see only that fancy error.

I agree there are cases when disabling backtracking is desirable. Maybe we could add something for this in version 7... Too tried right now to think properly, maybe on the weekend.

mrkkrp:
Going to keep this one as a feature request for the `commit` function.

mrkkrp:
As I'm working on version 8 now, I once again decided not to add `commit`. I'm closing this one, but #381 seems to be relevant, so maybe you'll be interested in that.

neongreen:
> I once again decided not to add commit

Why?

mrkkrp:
@neongreen See [here](https://github.com/mrkkrp/megaparsec/pull/361#issuecomment-546610846).

PR #361, opened 2019-07-18
[RFC] Memoization and backtracking frames

hesiod:
This is a request for comments on an experimental modification to Megaparsec that allows memoization of parsing results. There are still a number of rough edges around so this is not ready to be merged, but I would be grateful for feedback!
Overview of the changes:

1. A new combinator `memo :: m a -> m a` that memoizes a parser.
2. A new type `Memo e s m` that includes the actual memoization table and is passed around similar to the parsing state `State s`, but with some differences (e.g. when doing lookahead, we want to keep the memoization results but discard the resulting state).
3. The memoization table is implemented using a two-level map. The first level is an ordinary `IntMap` for looking up results based on token offset. The second level is a special data structure called `TagMap f` (open to suggestions on better names!). The idea behind that is that we want to associate parsers with their results at a specific token offset, and we need a way to identify different parsers. It is not enough to rely on a parser's type alone because different parsers can still have identical types, so we have to identify parsers by "value". This is not quite straightforward because a `ParsecT e s m a` is really just a CPS function. Fortunately, there is a way out: Using `StableName`s. They uniquely identify a value.
Essentially, a `TagMap f` associates a `StableName (f a)` with a value `f a`, but with a twist: We need to store parsers that return different result types in our memo table - otherwise, it would be useless. So the result type `a` is hidden using a combination of existential quantification and unsafe coercions to and from `Any`. The latter look scary, but they should be harmless since if two values have the same `StableName`, they are guaranteed to have the same type.
Note that we could completely remove `f` (which is `ParsecT e s m` in this case) and just use `a`. Why do we keep track of `f` at all then? The answer lies in type safety: By tracking the outer type `f`, we make sure that two memo tables with different type error/stream/underlying monad types are not compatible.
This version of the `TagMap` is the result of a number of rounds of completely reimplementing the memoization table, and I'm reasonably confident that this version is a good compromise.
4. Aside from memoization, this PR also includes a system to keep track of (nested) backtracking called "backtracking frames".
The main reason I implemented this was to save some space by discarding memo entries that are no longer required: If we leave the topmost frame, we are committed to that token offset, i.e. we will never look at tokens before the current offset, so we can remove all memo entries with an offset smaller than the current offset.
This has other benefits as well: By adding an extra `frameCommitted` bool to the frame data structure, we can easily add the `commit` operator (see #315). When starting a new frame (in `try`), we set the `frameCommitted` field to false. The new `commit` combinator just sets the current frame's `frameCommitted` field to true. When leaving the frame, we check whether the `frameCommitted` field is true. If it is, the frame was committed and we act as if we weren't backtracking at all. If it is not, we backtrack normally.

The performance benefits are promising. I've done some (limited) benchmarking and the results are as expected, i.e. there are no measurable performance penalties when you're not actually using memoization, but if you're memoizing often-reparsed parsers, parsing will be considerably faster (especially if your parsers are doing a lot of backtracking).

Open questions:
1. The `TagMap f` data structure is an ideal candidate for a `HashMap`. I didn't use it because it would incur an additional dependency on `unordered-containers`. Additionally, if the number of memoized results per offset is small (which is my expectation), the constant overhead of `HashMap`s could be significant compared to a simple linked list.
2. As noted in the documentation for `memo`, there are some possible issues regarding stateful monads as the underlying `ParsecT` monad. There might be further implications I'm not aware of.
3. I've initially added the backtracking frame system to the `Memo` type, but it might be better to move the frames to the main `State` type.
4. It's not quite clear what the correct behaviour of `withParsecT` should be. Currently, it maps the memo  entries using the supplied error type conversion function. Should `withParsecT` instead discard any memo entries? I'm not even sure what the intended use of `withParsecT` is.
5. It might be a better idea to add a new type `MemoParsecT e s m a` in addition to the standard `ParsecT e s m a` and only implement memoization for the former. This is possible since memoization should be transparent and `memo = id` is a valid implementation for `memo :: m a -> m a`, so the normal `ParsecT e s m a` can pretend to implement `memo` as well. Users could choose to utilize memoization by changing their parser type.
6. During debugging, I sprinkled some bang patterns here and there. Most of them had a positive impact when I inserted them, but some could certainly be removed again.

A possible future addition is incremental parsing. The idea is quite simple: Instead of discarding the memo table after parsing, it is kept around and reused when re-parsing the edited input. The only issue is determining which memo entries are affected by an edit operation and invalidating them. See https://ohmlang.github.io/pubs/sle2017/incremental-packrat-parsing.pdf for an in-depth introduction to the topic.
See https://blog.mattbierner.com/memoization-in-parse/ for a comparable implementation of memoization in a CPS parser combinator library. Some of the design choices I made were based on that article, including the explicit `memo` combinator and backtracking frames (called memoization window in the article).

### DIFF
--- a/Text/Megaparsec/Class.hs
+++ b/Text/Megaparsec/Class.hs
@@ -172,6 +172,54 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
 
   eof :: m ()
 
+  -- | Commit to the current parsing alternative. If we are currently inside a
+  -- backtracking frame (i.e. the current parser is running inside 'try'),
+  -- the innermost backtracking frame is marked as committed, effectively
+  -- undoing the effects of 'try' combinator.
+  --
+  -- For example, in the following parser, the parser commits to the first branch
+  -- after parsing “foo”. It then encounters the incorrect string “bar” and fails
+  -- even though it is run inside 'try'. The (matching) second branch is not considered
+  -- because the first branch already consumed some input.
+  --
+  -- >>> parseTest (try (string "foo" >> commit >> string "baz") <|> string "foobar") "foobar"
+  -- 1:4:
+  --   |
+  -- 1 | foobar
+  --   |    ^^^
+  -- unexpected "bar"
+  -- expecting "baz"
+  --
+  -- This combinator is very useful if you only want to backtrack on failure on the first part
+  -- of a parser. It allows you to regain some of the error reporting information lost
+  -- when backtracking.
+  --
+  -- @since 8.0.0
+  commit :: m ()
+
+  -- | The parser @'memo' p@ memoizes the inner parser @p@, i.e. the parser @p@
+  -- is run at most once for every possible location in the input stream.
+  --
+  -- There is no automatic memoization, you have to manually mark all parsers you
+  -- want to memoize.
+  -- Ideal candidates for memoization are parsers that are relatively expensive
+  -- to run (e.g. many alternatives or lots of token ingested before failure)
+  -- and parsers that are typically run multiple times during backtracking.
+  -- Note that memoization has associated space (storage of memoized results)
+  -- and time (memo lookup) overhead. In particular, if a lookup in the memoization
+  -- table is more expensive than running the parser, memoization is not a good
+  -- idea.
+  --
+  -- __N.B.__ The memoization is guaranteed to happen transparently (without affecting
+  -- parsing results) /iff/ either the underlying monad does not allow any side
+  -- effects observable during execution (e.g. 'IdentityT' or 'Control.Monad.Reader.ReaderT') or alternatively
+  -- the result of the memoized parser does not depend on any observable side effects.
+  -- This is because Megaparsec has no way to know that any given parser depends
+  -- on such side effects and might need to be re-run.
+  --
+  -- @since 8.0.0
+  memo :: m a -> m a
+
   -- | The parser @'token' test expected@ accepts a token @t@ with result
   -- @x@ when the function @test t@ returns @'Just' x@. @expected@ specifies
   -- the collection of expected items to report in error messages.
@@ -301,6 +349,8 @@ instance MonadParsec e s m => MonadParsec e s (L.StateT st m) where
   observing     (L.StateT m) = L.StateT $ \s ->
     fixs s <$> observing (m s)
   eof                        = lift eof
+  commit                     = lift commit
+  memo          (L.StateT m) = L.StateT $ memo . m
   token test mt              = lift (token test mt)
   tokens e ts                = lift (tokens e ts)
   takeWhileP l f             = lift (takeWhileP l f)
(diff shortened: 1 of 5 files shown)


hesiod:
I just realized there is a minor issue with the `tmap` function: The `unsafeCoerce` is harmless, but it is also useless because the `StableName` is for the entire `f a` value (not just the inner `a`), but we are changing the outer type in `tmap`. Therefore the `StableName` will never match (and lookup will always fail after transforming a `TagMap` using `tmap`) because the values are of type `f' a` but the `StableName`s are associated with (different) values of type `f a`.
This doesn't affect the main functionality of the memo table, it just means that the current implementation for changing the error type of a memo table in `withParsecT` is essentially useless.

mrkkrp:
I'll get to reviewing your stuff, just be patient. [Ormolu](https://github.com/tweag/ormolu) currently takes a lot of the energy I have for open source, and the rest is consumed by other matters.

hesiod:
@mrkkrp No worries, take your time - I'm in no rush to have this merged

mrkkrp:
I'm gonna get to reviewing this in the next couple weeks. Took almost 3 months :)

mrkkrp:
Hi @hesiod, sorry it took me so long to get to this.

Let's start from the beginning. What is the problem you're trying to solve? Can you give an example of a parser that greatly benefits from this change?

Regarding `commit`, it is true that there is a room for improvement for how different branches are combined and what is reported in the end, but I don't think that this solution is the best answer (it tries to solve a bit different problem related to performance though, where I'm personally dissatisfied with quality of error messages in some cases).

The reason is that `try` is already a pretty hard thing to work with, adding another combinator that can cancel its effects is going to just make the mental model of how a parser work even harder to understand.

Let's look at the example you give in the Haddocks. You could rewrite your parser like this to commit after `"foo"`:

```haskell
parseTest (string "foo" >> (string "baz" <|> string "bar")) "foobar"
```

So far, the need for `commit` in this form is not obvious imo.

Regarding `memo` and its supporting machinery. It looks like you're trying to get a particular parser you've have to  run very fast. The question here is whether or not this problem that you're trying to solve common enough to justify adding another primitive and whether or not it can be solved by writing the parser differently using what Megaparsec already provides. To answer these questions, as I said, I'd like to see an example that motivates this change.

mrkkrp:
Also, I found a significant enough (around 10%) performance penalty with this patch even when `memo` itself is not used.

[master-memory.txt](https://github.com/mrkkrp/megaparsec/files/3774972/master-memory.txt)
[master-speed.txt](https://github.com/mrkkrp/megaparsec/files/3774973/master-speed.txt)
[memo-memory.txt](https://github.com/mrkkrp/megaparsec/files/3774974/memo-memory.txt)
[memo-speed.txt](https://github.com/mrkkrp/megaparsec/files/3774975/memo-speed.txt)

This is `master` vs your (rebased on top of `master`) `memo` branch.

You can run the benchmarks yourself. The benchmark derivation is `benches.parsers-bench`. See instructions in [`HACKING.md`](https://github.com/mrkkrp/megaparsec/blob/master/HACKING.md).

mrkkrp:
Closing for now.

PR #362, opened 2019-07-20
Allow Stream to specify custom token width

hesiod:
Currently, the `errorItemLength` function implicitly assumes that every token spans exactly one character. That assumption is correct for the `Stream` instances for the basic string types (`String`/`Text`/`ByteString`) included in Megaparsec. It does not necessarily hold for custom token types however, in particular if using a separate lexer pass.

This commit adds a new `tokensLength` function that allows the `Stream` instance to give a custom width for a non-empty token stream.

In practice, parse errors concerning a token with a width greater than 1 are currently rendered incorrectly (only one caret is displayed below the offending token).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Megaparec 8.0.0
 
+* Added the `tokensLength` method to the `Stream` type class to improve
+  support for custom input streams.
+
 * Changed type signatures of `reachOffset` and `reachOffsetNoLine` methods
   of the `Stream` type class. Instead of three-tuple `reachOffset` now
   returns two-tuple because `SourcePos` is already contained in the returned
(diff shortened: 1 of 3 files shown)


mrkkrp:
I can see the problem, but I think this can be solved differently. It looks like `reachOffset` and `reachOffsetNoLine` functions in general already know width of tokens because they are responsible for incrementing `SourcePos`. Could we leverage that instead of adding a new method to `Stream`? I imagine we could use `reachOffsetNoLine` instead of `errorItemLength` to consume the tokens from the error message starting from current position `pst'` and then the difference in column positions will be the length of those tokens.

mrkkrp:
Although if we try to follow the path I described, we need a way to go from `NonEmpty (Token s)` to `s`, which does not exist. So maybe we should add the extra method of `Stream` after all.

PR #369, opened 2019-09-01
Drop remaining CPP, drop support for GHC 8.2.x

mrkkrp:
Currently blocked on dependencies. In particular: https://github.com/bos/aeson/pull/697. Let's wait a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,10 @@ cache:
 
 matrix:
   include:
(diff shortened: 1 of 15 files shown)


mrkkrp:
Since there are currently blockers preventing compilation of `cirterion` with GHC 8.8.1, I've decided not to add this version of GHC to the build matrix and add it later instead.

PR #379, opened 2019-10-23
Remove CPP from the library

facundominguez:
This PR splits `Text.Megaparsec.Internal` in three modules. 
* `Text.Megaparsec.Internal`: A module containing definitions that use the `Monad` instance of `ParsecT`.
* `Text.Megaparsec.Internal.Monad`: A module providing the `Monad` instance of `ParsecT`.
* `Text.Megaparsec.Internal.ParsecT`: A module providing the definition of `ParsecT` and all the definitions that don't depend on its `Monad` instance.

In addition, there are two variants of `Text.Megaparsec.Internal.Monad`, one for base earlier than 4.13, and one variant for base-4.13 and later. cabal chooses between the two depending on which version of base is available.


facundominguez:
Closing in favor of #380.

PR #380, opened 2019-10-24
Removes CPP from Text.Megaparsec.Internal (take 2)

facundominguez:
It splits Text.Megaparsec.Internal in two modules forming a cycle.

* Text.Megaparsec.Internal: mostly unchanged but without the Monad instance for ParsecT,
* Text.Megaparsec.Internal.Monad: containing the Monad instance for ParsecT.

In addition, there are two variants of Text.Megaparsec.Internal.Monad, one for base earlier than 4.13, and one variant for base-4.13 and later. cabal chooses between the two depending on which version of base is available.

### DIFF
--- a/Text/Megaparsec/Internal.hs
+++ b/Text/Megaparsec/Internal.hs
@@ -14,7 +14,6 @@
 --
 -- @since 6.5.0
 
-{-# LANGUAGE CPP                        #-}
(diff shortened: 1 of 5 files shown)


mrkkrp:
I think the overhead of more complex module organization is not worth it. As discussed, let's start thinking about a way to replace CPP with something better in GHC. In the meantime, I'm closing this.

PR #384, opened 2019-11-05
Allow registration of delayed parse errors

### LINKED ISSUE #359: Returning multiple parse errors in a ParseErrorBundle
In the changelog for v7.0, when `ParseErrorBundle` was introduced, it says:

> By default there will be only one `ParseError` in such a bundle, but it's possible to add more parse errors to a bundle manually.

But there doesn't seem to actually be any easy way to manually collect and report multiple non-fatal errors during parsing, which if I understand it is the whole point of `withRecovery`. Currently I'm doing the following:

```haskell
type ParsecST e s m = ParsecT e s (StateT [ParseError s e] m)

-- this is defined but private in Text.Megaparsec
initialPosState :: String -> s -> PosState s
initialPosState name s = PosState
  { pstateInput = s
  , pstateOffset = 0
  , pstateSourcePos = initialPos name
  , pstateTabWidth = defaultTabWidth
  , pstateLinePrefix = ""
  }

runParserST :: forall m s e a. Monad m =>
  ParsecST e s m a -> String -> s -> m (Either (ParseErrorBundle s e) a)
runParserST p n s = merge <$> runStateT (runParserT p n s) [] where
  merge :: (Either (ParseErrorBundle s e) a, [ParseError s e]) ->
      Either (ParseErrorBundle s e) a
  merge (a, []) = a
  merge (Right _, e : es) = Left (build es e [] (initialPosState n s))
  merge (Left (ParseErrorBundle (l :| ls) pos), e : es) =
    Left (build es e (l : ls) pos)
  build :: [ParseError s e] -> ParseError s e ->
    [ParseError s e] -> PosState s -> ParseErrorBundle s e
  build [] l ls = ParseErrorBundle (l :| ls)
  build (e : es) l ls = build es e (l : ls)

logNonFatalError :: Monad m => ParseError s e -> ParsecST e s m ()
logNonFatalError e = lift (modify (e :))
```

But this doesn't really look like user code to me. It would be much nicer if the monad supported a function like `logNonFatalError` for storing a parse error previously collected via `withRecovery`, which would then appear as additional errors in the bundle.

Am I going about this wrong? How do you envision people generating a `ParseErrorBundle` with multiple errors?

mrkkrp:
Close #359.

Megaparsec now supports registration of “delayed” parse errors. On lower level we added a new field called `stateParseErrors` to the `State` record. The type also had to change from `State s` to `State s e`. This field contains the list of registered `ParseErrors` that do not end parsing immediately but still will cause failure in the end if the list is not empty. Users are expected to register parse errors using the three functions: `registerParseError`, `registerFailure`, and `registerFancyFailure`. These functions are analogous to those without the `register` prefix, except that they have “delayed” effect.

TODO:

- [x] Running functions should fail if the list of delayed errors is not empty and the errors should be added to the parse error bundles.
- [x] Check effect on dependent packages.
- [x] Check how it affects performance if at all.
- [x] Clean up and make sure the Haddocks and other little details are correct.
- [x] Adjust/write tests.
- [x] Make all `deps` targets build. (I think only Idris remains now.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,16 @@ install:
   - travis_retry cabal new-update
 
 script:
-  - cabal new-build all --enable-tests --enable-benchmarks --flags=dev
-  - cabal new-test all --enable-tests --enable-benchmarks --flags=dev
+  # FIXME The recent changes are breaking and we need to release newer
+  # versions of hspec-megaparsec to make the tests work. I'll restore CI
+  # right after release of Megaparsec 8. For now we just test that it builds
+  # on CI and locally using Nix.
+  - cabal new-build megaparsec --enable-tests --enable-benchmarks --flags=dev
+  # - cabal new-test all --enable-tests --enable-benchmarks --flags=dev
   - cabal new-haddock megaparsec
-  - cabal new-haddock megaparsec-tests
+  # - cabal new-haddock megaparsec-tests
   - cabal new-sdist
-  - cd megaparsec-tests && cabal new-sdist
+  # - cd megaparsec-tests && cabal new-sdist
 
 notifications:
   email: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,20 @@
   new `parseError` method of `MonadParsec`. This method allows us to signal
   parse errors at a given offset without manipulating parser state manually.
 
+* Megaparsec now supports registration of “delayed” parse errors. On lower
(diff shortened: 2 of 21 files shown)


mrkkrp:
CC @digama0 @odanoburu. Here is the first rough draft. I'll finish it tomorrow. Let me know your thoughts.

digama0:
Looks good to me!

PR #398, opened 2020-01-23
Apply Ormolu

mrkkrp:
I decided not to check on CI for now, so the style is only recommended, but
not imposed (yet).


mrkkrp:
So, interestingly Haddock that comes with GHC 8.4.4 chokes on the output. One option is to wait till 8.10 is released and drop support for 8.4.

mrkkrp:
This is getting out of date with `master` and I'm going to delay formatting my code with Ormolu, because we have reasons to anticipate much more work on Ormolu sponsored by a big company in near future. It makes sense to apply Ormolu after that work is done.

Issue #388, opened 2019-11-26
Decouple error reporting from `Stream`


banacorn:
Many of the methods of `Stream`, like `reachOffset`,  `showTokens`, are related to printing pretty error messages. 

But it's a pain when it comes to implementing these methods for custom token streams.

Perhaps we should make error reporting a typeclass of its own:

```haskell
class Stream s => ReportError s where
```

So that people who have their own way of reporting error don't have to implement those methods.

mrkkrp:
It is like this because I wanted to do calculation of source position and fetching of line to display in one pass. I can imagine why you would want this, though.

Can you describe what exactly seems to be problematic about implementing `reachOffset` and `showTokens` in your case?

Also here is an example (granted, it's a very simple one): https://markkarpov.com/tutorial/megaparsec.html#working-with-custom-input-streams

banacorn:
I'm working on a parser that consumes custom token streams.
The tokens varied in "width" (like #370), some of them may even span across several lines.
This complicates the calculation of source positions and the offending line.

But the token streams I'm working with also happens to be equipped with source positions:

```haskell
-- | Run a lexer on a string and produce a lazy stream of tokens
runLexer
  :: forall tok.
     Lexer tok -- ^ lexer specification
  -> String -- ^ source file name (used in locations)
  -> String -- ^ source text
  -> TokenStream (L tok)
```

(here the [`L`](http://hackage.haskell.org/package/srcloc-0.5.1.2/docs/Data-Loc.html#t:L) is from `srcloc` for encompassing source locations)

In my case, I don't really need `reachOffset` or `showTokens`.
But it feels wrong not to implement these requirements (so i implemented them anyway)

ag-eitilt:
I'm having the same issue; I've written a lexer with a lot of error recovery, and while the output from that stage is usually not *too* changed from the input, there are still a lot of long, complex sequences compressed to a single token, a few tokens that get inserted from nothing, and a lot of the line breaks that have been removed.  With all that, even if "line" still had any meaning, I'd really rather just use the stage-1 tokens themselves in warning messages for the next parser pass (populating a data model), even if it means writing the printer myself.

[...]

PR #414, opened 2020-06-23
Add TokenStream and WrappedTokenStream

### LINKED ISSUE #413: Newtype for deriving stream type for lists
Would it be possible to make a stream newtype that makes it possible to derive (using derivingVia) a Stream type for newtype wrappers around list. The motivation is when you want to first use a lexer which collects a list of tokens and then parse. As it stands it seems you need to write your own Stream instance in those situations.

1Computer1:
This adds a new class `TokenStream` for streams which are made of tokens from a separate lexing step, as well as a `WrappedTokenStream` type for use in automatically deriving `Stream` instances from `TokenStream` instances.  

The interface of `TokenStream` and implementation of the `[]` instance are based on https://markkarpov.com/tutorial/megaparsec.html#working-with-custom-input-streams, as I do not really know any other way to implement a `Stream` instance for this.  

A problem currently is that doing `deriving via (WrappedTokenStream MyStream)` requires you to turn on `UndecidableInstances`. I haven't looked into a way to mitigate this. Maybe someone can come up with a better version of the `TokenStream` class or something better entirely.  

Nitpicks:
- More instances for other types like `Seq`.
- Bikeshedding method names.
- More detailed documentation.

This PR fixes #413.

### DIFF
--- a/Text/Megaparsec/Stream.hs
+++ b/Text/Megaparsec/Stream.hs
@@ -23,7 +23,10 @@
 --
 -- @since 6.0.0
 module Text.Megaparsec.Stream
-  ( Stream (..),
+  (
+    Stream (..),
+    TokenStream(..),
+    WrappedTokenStream(..)
   )
 where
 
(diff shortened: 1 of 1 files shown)


1Computer1:
I'll be honest, this is more of a result of not having #388. I'll play around to try and implement that instead.

PR #415, opened 2020-06-26
Decouple debug and error reporting methods from Stream

1Computer1:
This adds two new classes `StreamDebug` and `StreamError`, and moves these methods to them:
- `StreamDebug`: `showTokens`, `tokensLength`.
- `StreamError`: `reachOffset`, `reachOffsetNoLine`.

This allows users to implement `Stream` for streams where the token type is more complex and those methods may be difficult to implement.

This also adds `ListLikeStream` (name WIP) which lets you use `DerivingVia` for implementing `Stream` on newtypes over `[]` or `Seq`.

This PR addresses #388 and #413.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,6 +39,6 @@ jobs:
       - run: cabal v2-build all $CONFIG
       - run: cabal v2-test all $CONFIG
       - run: cabal v2-haddock megaparsec $CONFIG
-      - run: cabal v2-haddock megaparsec-tests $CONFIG
+      # - run: cabal v2-haddock megaparsec-tests $CONFIG
       - run: cabal v2-sdist
-      - run: pushd megaparsec-tests && cabal v2-sdist && popd
+      # - run: pushd megaparsec-tests && cabal v2-sdist && popd
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
-## Megaparsec 8.1.0
+## Megaparsec 9.0.0
+
+* Split the `Stream` type class. The methods `showTokens` and `tokensLength`
+  have been put into a separate type class `VisualStream`, while
+  `reachOffset` and `reachOffsetNoLine` are now in `TraversableStream`. This
+  should make defining `Stream` instances for custom streams easier.
+
+* Defined `Stream` instances for lists and `Seq`s.
 
 * Added the functions `hspace` and `hspace1` to the `Text.Megaparsec.Char`
   and `Text.Megaparsec.Byte` modules.
--- a/Text/Megaparsec.hs
+++ b/Text/Megaparsec.hs
@@ -203,7 +203,8 @@ parseMaybe p s =
 parseTest ::
   ( ShowErrorComponent e,
     Show a,
-    Stream s
+    VisualStream s,
+    TraversableStream s
   ) =>
   -- | Parser to run
   Parsec e s a ->
@@ -632,7 +633,7 @@ setInput s = updateParserState (\(State _ o pst de) -> State s o pst de)
 -- abuses the library.
 --
 -- @since 7.0.0
-getSourcePos :: MonadParsec e s m => m SourcePos
+getSourcePos :: (TraversableStream s, MonadParsec e s m) => m SourcePos
 getSourcePos = do
   st <- getParserState
   let pst = reachOffsetNoLine (stateOffset st) (statePosState st)
--- a/Text/Megaparsec/Byte.hs
+++ b/Text/Megaparsec/Byte.hs
@@ -93,7 +93,7 @@ space = void $ takeWhileP (Just "white space") isSpace
 
 -- | Like 'space', but does not accept newlines and carriage returns.
 --
--- @since 8.1.0
+-- @since 9.0.0
 hspace :: (MonadParsec e s m, Token s ~ Word8) => m ()
 hspace = void $ takeWhileP (Just "white space") isHSpace
 {-# INLINE hspace #-}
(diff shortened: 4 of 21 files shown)


1Computer1:
Ah, I should also mention that this PR is (obviously) a breaking change. Furthermore, I am unable to run the tests, because they depend on `hspec-megaparsec` which is broken with this change.
Is there something I should do about this, or should I let you handle it @mrkkrp?

mrkkrp:
I was having some time away from computer. I should be able to get to reviewing this weekend, or if not next weekend. I'll see about the tests.

mrkkrp:
This is not forgotten. I was sidetracked recently by switching CI of my all open source projects to GitHub actions. For now that work has higher priority. Right when I finish with that I'll get to this.

mrkkrp:
Sorry for a ridiculous delay.

Today I revisited #388 and looked at this PR. Let me first re-state the problem so that we have a concise problem statement and the solution (as I understand it) in one place.

#### The problem

Certain custom input streams are sufficiently different from `String` or `Text`-like streams so that `reachOffset` and `reachOffsetNoLine` are hard to implement correctly. `showTokens` and `tokensLength` may be a bit problematic too, AFAIU. This makes Megaparsec 8 hard to use with this type of input streams.

#### Motivation for the original desgin

IIRC the motivation was to lump together everything we might need to work with a `Stream` and so achieve simplicity in the sense that once you have defined an instance of `Stream` you are done—everything will just work™. Granted, this is not ideal.

#### This PR

The solution from this PR is a sensible one. If an input stream doesn't naturally support `reachOffset` or it's very hard to write, then let's just split the `Stream` type class and the provided functionality. Defining (a relatively simple) `Stream` instance gives the user most functionality, and then for fancier things the user may define `StreamDebug` and `StreamError`.

I see two problems:

1. I'm not entirely convinced by having two more classes. I think `StreamDebug` and `StreamError` can be merged. They both deals with the "visual" aspects i.e. how to represent and navigate input stream as something String-like. So the separation seems unnecessary. But this point is not very important.
2. The important point is that getting `SourcePos` (line number and column) must be possible for any input stream. This is just too basic a thing. It is strange to see functions like `indentBlock` to require the `StreamError` constraint, right? *"I'm not doing anything error-related, and yet here is this constraint!"* a demanding user might say, and they would be right.

#### What I propose

I seek a solution that would bright minimal disturbance in the Megaparsec ecosystem. To me it looks like it could be this:

*Do not implement what is too hard to implement.* If you have an input stream that is custom, you can just use dummy values for e.g. the `String` that `reachOffset` returns. I think obtaining `SourcePos` should still be feasible, it did not get more complex than before. In this case, naturally, some functionality of Megaparsec won't work exactly as intended (error printing), but if you have a custom stream you should be prepared to do things in a custom way sometimes. Write your own error printing function. If any of the helpers we have in Megaparsec are of use for that we can expose them (or maybe we could add a simpler error reporting function that just ignores that dummy offending line). This solution amounts to just extending the documentation, stating explicitly what can be "faked" and what functionality depends on which methods.

You might be thinking now that this solution is exactly the same as with splitting `Stream`, it's just more implicit and therefore worse. Not exactly. Combining the building of that `String` (for offending line) with calculation of `SourcePos` is reasonable for `String` and `Text`-like streams and it's faster that way. Separating `Stream` properly (in a way that the problem number 2, see above, is solved) means splitting `reachOffset` that currently does two different things (calculation of `String` and of `SourcePos`). Calculation of offending line would have to go to `StreamError`. I'd like to preserve the benefits that we already have with the combined solution and minimize breaking changes. This is why the solution with updating the docs seems better.

What do you think?

1Computer1:
I'm not sure if that is the best solution, since it requires maintaining documentation in multiple places. It would also make #413 a bit harder since we can't really come up with the methods for `[a]`.
One thing I thought of is generalizing `reachOffset` so that it doesn't return a `String` but maybe `Tokens s` instead?

1Computer1:
As for your points about the PR:

> I'm not entirely convinced by having two more classes. I think `StreamDebug` and `StreamError` can be merged. They both deals with the "visual" aspects i.e. how to represent and navigate input stream as something String-like. So the separation seems unnecessary. But this point is not very important.

I mostly agree on this, but `reachOffset` is harder to implement than just pretty printing.

> The important point is that getting SourcePos (line number and column) must be possible for any input stream. This is just too basic a thing.

Getting the line and column number of a certain token isn't hard, but getting those at an arbitrary offset is quite difficult. It could end up in the middle of a token perhaps.

> It is strange to see functions like indentBlock to require the StreamError constraint, right? "I'm not doing anything error-related, and yet here is this constraint!" a demanding user might say, and they would be right.

I think a bit of bikeshedding of the names and this won't be too bad.

As an aside, when I was working around `Stream`, I found that it was easier to make custom errors that point to a start and end in the source. Non-custom errors would just be converted to those custom errors too (though I couldn't find a way to convert all the different error types to it). Then, I would just read the source and pretty print directly (this handled errors spanning across lines too). This eliminated the need for `showTokens`, `tokensLength`, and `reachOffset` completely. So perhaps the solution is to attach an offset to `SourcePos` which the error printer could use? It could still be done in one traversal, I think.

mrkkrp:
OK, let's split it as you propose. What about changing the names like so:

* `StreamDebug` -> `VisualStream`
* `StreamError` -> `TraversableStream`

?

1Computer1:
That seems pretty reasonable, I'll change it to that.

mrkkrp:
What's going to happen now is that I'll find time (hopefully this weekend) and adjust everything that is necessary, then merge it.
I also have a couple ideas that I want to try.

When this is merged I'll proceed to release Megaparsec 9.0.

1Computer1:
> I also have a couple ideas that I want to try.

What might those be? Through the course of this issue, I have landed upon an alternative formulation of ParsecT and Stream myself (that is very breaking in relation of Megaparsec), but I wonder if I missed something. If you're interested, I can write a little example.

PR #427, opened 2020-10-12
Add Safe Haskell support

### LINKED ISSUE #425: Safe Haskell support
Currently, Megaparsec does not use the Safe or Trustworthy language extensions, which makes it more challenging to use from codebases that use Safe Haskell. I'm not sure if there are any unsafe exposed features; if not, it should be fine to mark modules as Safe or Trustworthy, but if so, it might be nice to make a module that re-exports a safe subset of Megaparsec.

mlugg:
Remove dependence on GeneralizedNewtypeDeriving and add Safe Haskell
extensions to every file. Closes #425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Megaparsec 9.0.1
+
+* Added [Safe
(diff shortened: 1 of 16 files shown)


mrkkrp:
Thanks!

PR #434, opened 2020-12-11
Text.Megaparsec.Char.charNegate

jamesdbrock:
Hi Mark, here's a combinator which does the same thing as regex character class negation.

* I've found this combinator to be generally useful.
* It doesn't seem to be implemented in any library that I can see.
* It not obvious how to write it (not obvious to me anyway, I had to give it a think).

What do you think? If you like it, I'll write some tests.

### DIFF
--- a/Text/Megaparsec/Char.hs
+++ b/Text/Megaparsec/Char.hs
@@ -52,6 +52,7 @@ module Text.Megaparsec.Char
     -- * Single character
     char,
     char',
+    charNegate,
 
     -- * Sequence of characters
     string,
@@ -308,6 +309,20 @@ char' c =
       char (toTitle c)
     ]
 {-# INLINE char' #-}
+
+-- | Combinator to “negate” a single-character parser.
+--
+-- Similar to “negation” of character classes in regex. Will produce a parser
+-- which will parse any single character /except/ the characters allowed
+-- by the input parser.
+--
+-- Parse any single character /except/ a lowercase letter or underscore:
+--
+-- >>> parseTest (charNegate $ lowerChar <|> char '_') "2"
+-- '2'
+charNegate :: (MonadParsec e s m, Token s ~ Char) => m (Token s) -> m (Token s)
+charNegate p = notFollowedBy p *> anySingle
+{-# INLINE charNegate #-}
 
 ----------------------------------------------------------------------------
 -- Helpers


mrkkrp:
This belongs to the `Text.Megaparsec` module because it is not specific to characters. Just like e.g. `anySingle`. I think the name of the function also should reflect this. Maybe it could be named `anySingleBut`?

On one hand it I welcome the addition; on the other hand though, I'd never use the function myself because constructing a custom predicate and then using `satisfy` will be faster. One could hope that `anySingleBut` would produce better error messages, but since it uses `notFollowedBy`, it's also not the case.

Question for you: why would you use this idiom over using `satisfy` with a predicate?

jamesdbrock:
> I'd never use the function myself because constructing a custom predicate and then using `satisfy` will be faster... why would you use this idiom over using `satisfy` with a predicate?

Well because I wanted to be able to write for example

```Haskell
nonProsaic = charNegate $ letterChar <|> punctuationChar
```

which pleases me as a nice parser composition, and which I could write without importing anything else.

But now I have looked more closely and I realize that if I import [`Data.Char`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Char.html) then I can do

```Haskell
nonProsaic = satisfy $ \c -> not $ isLetter c || isPunctuation c
```

which I think is what you had in mind, and which is better for all the reasons you gave. So yeah, it's probably not a good idea to merge this PR.

On a related subject, all of the links to `Data.Char` in the [`Text.Megaparsec.Char` 9.0.1](https://hackage.haskell.org/package/megaparsec-9.0.1/docs/Text-Megaparsec-Char.html) Haddocks are currently pointing to a version of base which doesn't exist in Hackage.

jamesdbrock:
Maybe it would be worth it to include a sentence of documentation in `Text.Megaparsec.Char` which explains how to do regex-style “character class negation.”

mrkkrp:
> On a related subject, all of the links to Data.Char in the Text.Megaparsec.Char 9.0.1 Haddocks are currently pointing to a version of base which doesn't exist in Hackage.

I had to generate the docs manually for some reason, and that's the version they got linked to. It is a shame that Hackage maintainers do not upload docs for built-in packages like `base`. I do not know what I could do about this.

I'll happily add a sentence about preferring `satisfy` for writing complex parsers that consume a single character. I'm not sure where to put it though. In the docs of `satisfy` itself? Do you have an option?

PR #439, opened 2021-01-08
Fix Haddock link to Text.Megaparsec.Char.space1

cdepillabout:
The Haddock links for `space1` don't currently go anywhere.  This PR fixes these links:

https://hackage.haskell.org/package/megaparsec-9.0.1/docs/Text-Megaparsec-Char-Lexer.html

My guess is that `Text.Megaparsec.Char.Lexer` was imported qualified as `C`, so the links used to work, but that import has since been removed.

### DIFF
--- a/Text/Megaparsec/Lexer.hs
+++ b/Text/Megaparsec/Lexer.hs
@@ -36,11 +36,11 @@ import Text.Megaparsec.Common
 -- @spaceConsumer@ in documentation, usually it means that something like
 -- 'space' is expected there).
 --
--- @sc@ is used to parse blocks of space characters. You can use 'C.space1'
+-- @sc@ is used to parse blocks of space characters. You can use 'Text.Megaparsec.Char.space1'
 -- from "Text.Megaparsec.Char" for this purpose as well as your own parser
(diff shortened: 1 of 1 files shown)


Text/Megaparsec/Lexer.hs
```diff
@@ -36,11 +36,11 @@ import Text.Megaparsec.Common
 -- @spaceConsumer@ in documentation, usually it means that something like
 -- 'space' is expected there).
 --
--- @sc@ is used to parse blocks of space characters. You can use 'C.space1'
+-- @sc@ is used to parse blocks of space characters. You can use 'Text.Megaparsec.Char.space1'
 -- from "Text.Megaparsec.Char" for this purpose as well as your own parser
 -- (if you don't want to automatically consume newlines, for example). Make
 -- sure the parser does not succeed on empty input though. In earlier
--- version 'C.spaceChar' was recommended, but now parsers based on
+-- version 'spaceChar' was recommended, but now parsers based on
```
cdepillabout, line 43:
```suggestion
-- version 'Text.Megaparsec.Char.spaceChar' was recommended, but now parsers based on
```

mrkkrp:
Thanks!

PR #460, opened 2021-07-22
Machine number parsers (Word16/32/64, both endians)

### LINKED ISSUE #459: Parsers for Word16, Word32, Word64 (BE and LE)
(Half issue, half PR.) I'm doing some binary format parsing and was surprised to find that there's not much on Hoogle for obtaining unsigned integers like Word16. I ended up having to use some monadic Get computation from `binary`, which also had me converting my strict ByteString to a lazy one. Wasn't happy, wrote a helper library for Megaparsec to parse `[Word8]` to the various other Word types.

Would this be useful as an extra library to Megaparsec, or would you maybe consider adding it directly? I feel like it's handy for binary format parsers. Attoparsec has a simple package `attoparsec-binary`, but it's long been unmaintained.

The aim being to provide a handful of parsers for doing things like:

```haskell
userParser = do
...
  value1 :: Word32 <- anyW32LE
  value2 :: Word16 <- anyW16BE
  value3 :: Word64 <- anyLE
...
```

And knowing that it works on any stream of `Word8` tokens with OK performance. Is this the sort of thing other Megaparsec users might find useful?

---

I have an initial library at https://github.com/raehik/megaparsec-binary . It's <100 SLOC and has no additional dependencies. The API is not well thought-out (function names etc).

raehik:
Close #459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Megaparsec 9.2.0
+
+* Added parsers for binary representations (little/big endian) of numbers in
+  `Text.Megaparsec.Byte.Binary`.
+
 ## Megaparsec 9.1.0
 
 * Added `dbg'` in `Text.Megaparsec.Debug` for debugging parsers that have
--- a/Text/Megaparsec/Byte/Binary.hs
+++ b/Text/Megaparsec/Byte/Binary.hs
@@ -1,0 +1,192 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE Safe #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- |
+-- Module      :  Text.Megaparsec.Byte.Binary
+-- Copyright   :  © 2021–present Megaparsec contributors
+-- License     :  FreeBSD
+--
+-- Maintainer  :  Mark Karpov <markkarpov92@gmail.com>
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- Binary-format number parsers.
+--
+-- @since 9.2.0
+module Text.Megaparsec.Byte.Binary
+  ( -- * Generic parsers
+    BinaryChunk (..),
+    anyLE,
+    anyBE,
+
+    -- * Parsing unsigned values
+    word8,
+    word16le,
+    word16be,
+    word32le,
+    word32be,
+    word64le,
+    word64be,
+
+    -- * Parsing signed values
+    int8,
+    int16le,
+    int16be,
+    int32le,
+    int32be,
+    int64le,
+    int64be,
+  )
+where
+
+import Data.Bits
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Lazy as BL
+import Data.Int
+import Data.Word
+import Text.Megaparsec
+
+-- | Data types that can be converted to little- or big- endian numbers.
+class BinaryChunk chunk where
+  convertChunkBE :: (Bits a, Num a) => chunk -> a
+  convertChunkLE :: (Bits a, Num a) => chunk -> a
+
+instance BinaryChunk B.ByteString where
+  convertChunkBE = B.foldl' go 0
+    where
+      go acc byte = (acc `unsafeShiftL` 8) .|. fromIntegral byte
+  convertChunkLE = B.foldl' go 0
+    where
+      go acc byte = (acc .|. fromIntegral byte) `rotateR` 8
+
+instance BinaryChunk BL.ByteString where
+  convertChunkBE = BL.foldl' go 0
+    where
+      go acc byte = (acc `unsafeShiftL` 8) .|. fromIntegral byte
+  convertChunkLE = BL.foldl' go 0
+    where
+      go acc byte = (acc .|. fromIntegral byte) `rotateR` 8
+
+----------------------------------------------------------------------------
+-- Generic parsers
+
+-- | Parse a little-endian number.
+--
+-- You may wish to call this with a visible type application:
+--
+-- > number <- anyLE (Just "little-endian 32 bit word") @Word32
+anyLE ::
+  forall a e s m.
+  (MonadParsec e s m, FiniteBits a, Num a, BinaryChunk (Tokens s)) =>
+  -- | Label, if any
+  Maybe String ->
+  m a
+anyLE mlabel = convertChunkLE <$> takeP mlabel (finiteByteSize @a)
+{-# INLINE anyLE #-}
+
+-- | Parse a big-endian number.
+--
+-- You may wish to call this with a visible type application:
+--
+-- > number <- anyBE (Just "big-endian 32 bit word") @Word32
+anyBE ::
+  forall a e s m.
+  (MonadParsec e s m, FiniteBits a, Num a, BinaryChunk (Tokens s)) =>
+  -- | Label, if any
+  Maybe String ->
+  m a
+anyBE mlabel = convertChunkBE <$> takeP mlabel (finiteByteSize @a)
+{-# INLINE anyBE #-}
+
+--------------------------------------------------------------------------------
+-- Parsing unsigned values
+
+-- | Parse a 'Word8'.
+word8 :: (MonadParsec e s m, BinaryChunk (Tokens s)) => m Word8
+word8 = anyBE (Just "8 bit word")
+{-# INLINE word8 #-}
+
+-- | Parse a little-endian 'Word16'.
+word16le :: (MonadParsec e s m, BinaryChunk (Tokens s)) => m Word16
+word16le = anyLE (Just "little-endian 16 bit word")
+{-# INLINE word16le #-}
+
+-- | Parse a big-endian 'Word16'.
+word16be :: (MonadParsec e s m, BinaryChunk (Tokens s)) => m Word16
+word16be = anyBE (Just "big-endian 16 bit word")
+{-# INLINE word16be #-}
+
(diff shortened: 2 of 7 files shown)


raehik:
I need to write tests (I get to use QuickCheck for the first time!), and the API and naming is certainly up for debate. Not really happy with calling them "machine numbers", but unsure what else to call them.

raehik:
Would appreciate any comments on the implementation, I used some concepts that I've only recently learned about.

raehik:
The implementation could potentially be made (lots?) faster by using the `ByteString` directly (currently I convert to `[Word8]) and `unsafeIndex` to grab the individual bytes (also omitting the fold). I would look into that, but it would seem I'd have to write it twice, for lazy and strict ByteStrings.

Text/Megaparsec/Byte/MachineInts.hs
```diff
@@ -0,0 +1,110 @@
+-- TODO debug
```
mrkkrp, line 1: I think we could call this module `Text.Megaparsec.Byte.Binary`.

Text/Megaparsec/Byte/MachineInts.hs
```diff
@@ -0,0 +1,110 @@
...
+
+import Data.Bits
+-- TODO debug below
+import qualified Data.ByteString as BS
+import Data.Maybe (fromMaybe)
+import Data.Proxy (Proxy (..))
+import Data.Void (Void)
+import Data.Word
+import Text.Megaparsec
+import Text.Megaparsec.Byte.MachineInts.Internal
```
mrkkrp, line 18: I think it is better to merge the two modules.

Text/Megaparsec/Byte/MachineInts.hs
```diff
@@ -0,0 +1,110 @@
+-- TODO debug
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | "Machine" integer parsers.
+module Text.Megaparsec.Byte.MachineInts where
```
mrkkrp, line 8: Please add an explicit export list with all public functions.

Text/Megaparsec/Byte/MachineInts.hs
```diff
@@ -0,0 +1,110 @@
...
+
+anyW64BE ::
+  forall e s m.
+  (MonadParsec e s m, Token s ~ Word8) =>
+  m Word64
+anyW64BE = anyBE
+
+--------------------------------------------------------------------------------
+
+parseTest' :: Show a => Parsec Void BS.ByteString a -> BS.ByteString -> IO ()
```
mrkkrp, line 92: Obviously, these testing functions should be removed.

megaparsec-tests/tests/Text/Megaparsec/Byte/MachineIntsSpec.hs
```diff
@@ -0,0 +1,56 @@
...
+    it "parses a little-endian Word32" $ do
+      let p = anyLE @Word32 <* eof
+          (s, _, valueLE) = exW32
+      prs p s `shouldParse` valueLE
+      prs' p s `succeedsLeaving` ""
+
+----------------------------------------------------------------------------
+-- Helpers
+
+prs ::
```
mrkkrp, line 32: Why do you re-define these functions here? They are available in `Test.Hspec.Megaparsec.AdHoc`.

megaparsec-tests/tests/Text/Megaparsec/Byte/MachineIntsSpec.hs
```diff
@@ -0,0 +1,56 @@
...
+type Parser = Parsec Void ByteString
+
+spec :: Spec
+spec = do
+  describe "anyLE" $ do
+    it "parses a little-endian Word32" $ do
+      let p = anyLE @Word32 <* eof
+          (s, _, valueLE) = exW32
+      prs p s `shouldParse` valueLE
+      prs' p s `succeedsLeaving` ""
```
mrkkrp, line 27: In this case `succeedsLeaving` is redundant because you have `eof` in your parser. `shouldParser` will never succeed unless it consumes all input. Indeed, QuickCheck could be useful here. A typical test could be:

* Generate a number.
* Serialize it.
* Parse it and compare with the originally generated number.

mrkkrp (commented):
I don't know what's the fastest way to parse, I'd need to look at it when I have more time. I suspect that going through `ByteString`s is not possible unless you add more constraints to the functions because there are different stream types that all have `Token s ~ Word8`, i.e. the `Token` type function is not injective. Your stream could be as well just `[Word8]`.

Text/Megaparsec/Byte/MachineInts.hs
raehik: That's what I called it at first, then stopped and looked. Thinking about it again it makes sense as an abbreviation for "binary-format numbers". Will change

Text/Megaparsec/Byte/MachineInts.hs
raehik: Yep, this was primarily to separate Megaparsec functions from utility ones. No need if it's actually in Megaparsec.

megaparsec-tests/tests/Text/Megaparsec/Byte/MachineIntsSpec.hs
raehik: Oops, copy-pasted wrong. Thanks.

raehik:
Thanks for the review! I'm not in a rush for this, so I'll get to it later this weekend or next week. Kind of want to see if there's any real difference between the `[Word8]` approach and the more boilerplate-y `ByteString.unsafeIndex` approach used in `binary`, `cereal` (out of curiosity more than real concerns, since I feel this should be fine for Megaparsec)

megaparsec-tests/tests/Text/Megaparsec/Byte/MachineIntsSpec.hs
raehik: Wait, no, `Test.Hspec.Megaparsec.AdHoc` defines a parser over a `String` stream. `ByteSpec` and `Byte.LexerSpec` both redefine it for `ByteString` instead, which I need too. What should I do here? Leave at copy-paste? ByteString Parser module? Stream type typevar in `Parser`?

Text/Megaparsec/Byte/MachineInts.hs
raehik: (This helps me avoid weird GHCi issues due to typeclasses (I think) when testing. Will remove & squash history before committing.)

raehik:
I lied, serializing was easy due to ByteString builders. I'll clean up the impl module and write some more tests later.

raehik:
Still want to look at performance numbers at some point, but they don't matter enough to pause this PR.

I've resolved everything except the redefined `prs`, `Parser` utils. I've added tests for the things that matter - think however that testing the specific parsers `anyW16LE` is overkill, your thoughts?

raehik:
Didn't realise I left this as draft. Ready for review.

mrkkrp:
I’m on vacation right now, I may not have time to look at it for a while.

raehik:
No rush at all -- wishing you a relaxing/fun holiday! :)

raehik:
Quick update. I've been using this a lot for some binary reverse engineering and it's very handy. No performance metrics, but it's got no problem working in a huge sum data parser & is extremely convenient (if curious: [big parser](https://github.com/raehik/gtvm-hs/blob/3fd009c68f18f47dc84cc9c6978f7c561d3bc062/src/GTVM/SCP/Parse.hs#L63), [auto LE/BE wrappers](https://github.com/raehik/gtvm-hs/blob/cdcdba2b80ddb45e19df9f7159dae876acfa98a4/src/GTVM/Common/Binary/Parse.hs#L29)).

I originally wanted this to match a feature in a Rust parser combinator library. It's enabled me to use Megaparsec as an extremely convenient early-stage binary parser. Though error messages are quite broken: wonder if `Show ByteString` is more at fault there.

mrkkrp:
I started looking today. I think we need parsers for both words and ints in the module. There are a few things related to performance too. It is not a good idea to go via lists, so I had to introduce a type class that would use `foldl'` that directly folds over strict and lazy byte strings. I believe I finished refactoring the new module, but I haven't looked at the tests yet and I'll need to add benchmarks. No estimates when this will be done.

mrkkrp:
I'd say it is ready to be merged. Regarding broken error messages, perhaps we could prove an option to display a stream of hex values (bytes) somehow and then print a pointer there, what do you think?

mrkkrp (approved):

raehik:
Fantastic! Much better solution for working with any bytestream-like stream. Thank you so much for reworking it so superbly.

> Regarding broken error messages, perhaps we could provide an option to display a stream of hex values (bytes) somehow and then print a pointer there, what do you think?

I would find that very useful for working with binary formats, yeah. I ended up making a newtype wrapper over strict `ByteString`s for `00 80 FE FF` etc parse-printing. Being able to print error messages that way too would be great (currently I rely on the byte position provided), but I'm not sure about I'd implement it. Many parsers work on `ByteString`s but are actually text (or primarily), so printing all `ByteString`s as hex byte streams wouldn't be ideal.

mrkkrp:
Perhaps we could somehow allow the user to opt in for hexadecimal input stream in error messages...

raehik:
That makes sense to me. Default to text printing, but allow swapping to hex byte printing for streams that can be interpreted as bytestreams. I don't know much about the error system in Megaparsec, so I was wondering if it'd be a lot of work to handle pretty errors etc. nicely. I would definitely work on this if you think it's viable!

mrkkrp:
I think I'll merge this and open a separate issue for the binary pretty-printing.

PR #463, opened 2021-08-11
Megaparsec.Class.MonadParsec: token: upd docs

### LINKED ISSUE #462: Please, improve the documentation of token{,s}
If to compare what `token{,s}` are semantically & explanation of them - the semantics seem simpler than the explanation.

Anton-Latukha:
Close #462.

Applying simplification to structure while keeping the important information.

### DIFF
--- a/Text/Megaparsec/Class.hs
+++ b/Text/Megaparsec/Class.hs
@@ -153,14 +153,14 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
   -- | This parser only succeeds at the end of input.
   eof :: m ()
 
-  -- | The parser @'token' test expected@ accepts a token @t@ with result
-  -- @x@ when the function @test t@ returns @'Just' x@. @expected@ specifies
-  -- the collection of expected items to report in error messages.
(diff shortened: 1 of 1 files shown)


mrkkrp:
Thanks!

I don't think we should abandon the rather common practice of showing an invocation of a function with named arguments and referring to those arguments in the explanation. The font Haddock uses by default is perhaps not very different from normal text (although I'd say it is different enough), but that's an issue to report on the Haddock issue tracker. Also note that the kind of list you tried to introduce will not be displayed by Haddock as such, everything will be joined together in one line. The alternatives are: either use Haddock's special syntax for lists, or rephrase to make it a normal sentence (I took the liberty of doing that).

Anton-Latukha:
Thank you.

Yes, indeed, parsing of Haddock has a number of caveats, on top of that the editors (Emacs in my case) & Hackage have differences in the representation of Haddock.

Probably the contributors of Haddocks should reference generated HTML (which I going to do further).

---

> The alternatives are: either use Haddock's special syntax for lists, or rephrase to make it a normal sentence (I took the liberty of doing that).

Fully agree, documentation process inside constraints of merge requests  - works efficiently only this way.

PR #466, opened 2021-08-31
Small doc update

Anton-Latukha:
During study/work seeing what is missing from official docs (haddocks) or what is hard to understand from them, and making the patch.

### DIFF
--- a/Text/Megaparsec/Class.hs
+++ b/Text/Megaparsec/Class.hs
@@ -205,11 +205,9 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
     m (Tokens s)
 
   -- | Parse /zero/ or more tokens for which the supplied predicate holds.
-  -- Try to use this as much as possible because for many streams the
+  -- Try to use this as much as possible because for many streams this
   -- combinator is much faster than parsers built with
   -- 'Control.Monad.Combinators.many' and 'Text.Megaparsec.satisfy'.
-  --
-  -- The following equations should clarify the behavior:
   --
   -- > takeWhileP (Just "foo") f = many (satisfy f <?> "foo")
   -- > takeWhileP Nothing      f = many (satisfy f)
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Well, during learning - that is all that I wanted to explicitly add.

To give a recommendation not only for `takeWhile1P` but direct in `takeWhileP` also.

mrkkrp:
But isn't this already mentioned in at least three places?

* In the beginning of the section about primitive combinators (mentioning both `takeWhileP` and `takeWhile1P` as well as other combinators).
* In the official tutorial: https://markkarpov.com/tutorial/megaparsec.html#writing-efficient-parsers.
* `takeWhileP` explicitly talks about the performance advantages of `takeWhileP` over parsers built with `many` and `satisfy`.

Since `takeWhile1P` is described in terms of `takeWhileP`:

> Similar to `takeWhileP`, but fails if it can't parse at least one token.

I'd assume that it is already clear that it is faster than parsers defined in terms of `some` and `satisfy`.

Anton-Latukha:
It is mentioned everywhere - except in the documents to the function.

The type class mentions it. `takeWhileP` was said to be "faster", but not said what is a slower analog of it is which to replace it with. The person is required to assume/infer it implicitly.

That is why I mentioned that if a person arrived to `some (satisfy f)` - it is `takeWhileP`.

To see that In the official tutorial person needs to read through just 11 426 -s words.

That people read it in the `takeWhile1P` description - does not mean that it also applies to `takeWhileP` - so to understand that - people need to read & understand through the code.

`takeWhileP` information can be removed information from typeclass description & from tutorial, if it is in the documentation of the `takeWhileP`  definition - that is where people look most directly & most often & expect to find information on the function. People also are reading docs from inside of the IDEs - that shows ... you know.

Anton-Latukha:
> I'd assume that it is already clear that it is faster than parsers defined

That is obvious to the parser creator. & those who used the old parser combinators.

But the person that learns parsers for the first time - that information is obscure & needs to be inferred & then checked.

Anton-Latukha:
Thank you :)

PR #485, opened 2022-08-18
Request minimal base 4.15

mrkkrp:
This corresponds to support for the latest 3 GHC versions.

### DIFF
--- a/megaparsec-tests/megaparsec-tests.cabal
+++ b/megaparsec-tests/megaparsec-tests.cabal
@@ -25,7 +25,7 @@ library
     default-language: Haskell2010
     build-depends:
         QuickCheck >=2.10 && <2.15,
-        base >=4.13 && <5.0,
+        base >=4.15 && <5.0,
         bytestring >=0.2 && <0.12,
         containers >=0.5 && <0.7,
         hspec >=2.0 && <3.0,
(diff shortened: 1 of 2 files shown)


sternenseemann:
I would like to request that this change is reverted. It doesn't seem to be necessary, as `megaparsec` 9.2.2 can be built base 4.14  and 4.13 and we would like to build packages with the most recent version of megaparsec also on older compilers where appropriate!

mrkkrp:
It is true that `megaparsec` builds with older bases and GHCs. This change (applied systematically to all packages I maintain) reflects my decision to only officially support 3 latest versions of GHCs. There is a distinction between "happens to build" and "officially supported", e.g. there is no CI job to build with GHC older than 9.0. I'm not willing to put effort to try to maintain a wider compatibility window. I do put effort though to make sure newest GHCs are supported as soon as possible. The same applies to other projects, e.g. Ormolu—even though they might work with an older version of GHC/base, it is phased out so that at any given moment only 3 versions of GHC are officially supported.

The reasoning here is simple. I do unpaid volunteer work here maintaining about 20 packages and have been doing it for about 7 years now in a reliable fashion. I haven't abandoned a single package and I'm updating packages to support latest software within 1 month typically. Anything that allows me to worry less and use less conditional compilation and CPP I will take. Some packages are easier to make work with a wide range of GHCs, some are harder. But, when this decision (3 latest GHC versions to support) is applied systematically, you see commits like this, which, isolated, may seem odd.

Some people may be willing to help co-maintain a package, but then suddenly they are not around to do actual work when it is needed. It is always me who is alert and updating stuff, so I do consider myself the sole maintainer. I cannot count on others, therefore I'm going to do it the way I deem reasonable.

PR #489, opened 2022-10-02
Add a type class for parser monads with debugging

### LINKED ISSUE #488: Type class for parser monads with debugging
As it is mentioned in [docs](https://hackage.haskell.org/package/megaparsec-9.2.2/docs/Text-Megaparsec-Debug.html#v:dbg)

> it's not possible to lift this function into some monad transformers without introducing surprising behavior (e.g. unexpected state backtracking) or adding otherwise redundant constraints (e.g. [Show](https://hackage.haskell.org/package/base-4.16.3.0/docs/Text-Show.html#t:Show) instance for state), so this helper is only available for [ParsecT](https://hackage.haskell.org/package/megaparsec-9.2.2/docs/Text-Megaparsec-Internal.html#t:ParsecT) monad, not any instance of [MonadParsec](https://hackage.haskell.org/package/megaparsec-9.2.2/docs/Text-Megaparsec.html#v:MonadParsec) in general

However, there are cases, when it would be very useful. For example for debugging parser with some state:
```hs
xCounter :: (MonadParsec Void String m, MonadState Depth m) => m Int
``` 
I propose to add special class (something like `MonadParsecDbg`) and instances for it with "redudant constraints", i. e. `Show` instance for state and writer's log type. This also gives us very useful state logging in debug mode (if we consider composed transformer `m` as a parser with state, it seems very natural for me, that it is printed along with parser state). 

Some [example](https://gist.github.com/Lev135/c2b7d9de30a6476e9d7d4d492dbfaca5) of how it can be used

Suggested solution:
```hs
class Monad m => MonadParsecDbg m where
  dbg :: Show a => String -> m a -> m a

instance (VisualStream s, ShowErrorComponent e) => MonadParsecDbg (ParsecT e s m) where
  dbg = Text.Megaparsec.Debug.dbg

instance (Show s, MonadParsecDbg m) => MonadParsecDbg (StateT s m) where
  dbg str sma = StateT $ \s ->
    dbg str $ runStateT sma s

instance (Show w, Monoid w, MonadParsecDbg m) => MonadParsecDbg (WriterT w m) where
  dbg str wma = WriterT $ dbg str $ runWriterT wma

instance (MonadParsecDbg m) => MonadParsecDbg (ReaderT e m) where
  dbg str rma = ReaderT $ \e -> dbg str $ runReaderT rma e
```
Maybe it would be better to add `MonadParsec` constraint for class and `e`, `s` parameters. Here I've simplified it for to make it more readable.

Lev135:
Close #488.

`MonadParsecDbg` class and instances for `MonadParsecT`, `ReaderT`, `WriterT`, `StateT` and `RWST` added.

There are some difficulties with testing these changes, because they use `trace` and thefore can not be tested by `hspec`. To work around this, I've added separate `test-debug` executablle into `megaparsec-tests` for running some parsers manually and checking produced ouput with my eyes. Maybe, this is not a good solution, but I have not found any better.

Futhermore, I have no `nix-shell` and it will be not very easy to run tests (I'm working on Windows). For instance, the `cabal test all` failed with `commitBuffer: invalid argument (invalid character)`. I don't think that my changes should break anything, however.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   negative, similar to when `n == 0`. [Issue
   497](https://github.com/mrkkrp/megaparsec/issues/497).
 
+* Added the `MonadParsecDbg` type class in `Text.Megaparsec.Debug`. The type
+  class allows us to use `dbg` in MTL monad transformers. [Issue
+  488](https://github.com/mrkkrp/megaparsec/issues/488)
+
 ## Megaparsec 9.2.2
 
 * Fixed a space leak in the implementations of the `reachOffset` and
--- a/Text/Megaparsec/Debug.hs
+++ b/Text/Megaparsec/Debug.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE Unsafe #-}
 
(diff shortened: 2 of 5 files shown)


mrkkrp:
I will review this soon. I think there could be a way to capture `stderr` output and examine it automatically, i.e. as an ordinary unit test without introducing an extra executable. I will fiddle with it when I have enough time.

mrkkrp:
The general gist of this PR is good. I was not happy with the testing approach because it should be completely automatic. Today I added `shouldStderr` to solve that and started adjusting `DebugSpec`. `debug/Test.hs` really belongs there. If you'd like you can continue converting the remaining part or otherwise I will soon find more time to finish it up.

mrkkrp (approved):

mrkkrp:
Thanks!

Issue #492, opened 2022-10-13
Unexpected input-sharing behaviour of parsers


EggBaconAndSpam:
I finally managed to hunt down a memory leak which arose from the following arguably surprising behaviour:

Running either of these parsers and holding on to their results will keep the entire input in memory:
```
import Data.Text (Text)
import qualified Data.Text.Lazy as TL

testSharingStrict :: Parsec Void Text Text
testSharingStrict = takeP Nothing 1

testSharingLazy :: Parsec Void TL.Text TL.Text
testSharingLazy =
  fmap TL.concat . many . try $ takeP Nothing 1 <* takeP Nothing (TL.defaultChunkSize - 1)
```

This is due to the sharing semantics of both strict and lazy text. (I believe the same applies to strict and lazy bytestrings.)

Having been bitten by this I'm struggling to come up with a scenario where we actually gain anything from sharing memory with the input stream. In the strict case that always keeps the entire input in memory, while in the lazy case we at least keep a whole chunk (of 16K) for each 'output' (which may be much smaller than that).

It seems to me we could get rid of the input-sharing behaviour be adding a few `copy :: Text -> Text` in the right places:
```
instance Stream T.Text where
  ...
  takeN_ n s
    | n <= 0 = Just (T.empty, s)
    | T.null s = Nothing
    | otherwise = Just (first T.copy $ T.splitAt n s)
  takeWhile_ = first T.copy $ T.span
```

Why is that a bad idea, and/or what am I missing?

EggBaconAndSpam:
One way to get rid of this foot-gun:
- add `copy` to the right places of `instance Stream Text` etc. to _not_ share default
- add a `Shared` newtype wrapper s.t. `instance Stream (Shared Text)` retains the current sharing behaviour

mrkkrp:
This is a valid concern, thanks for bringing this up. I think I originally wanted combinators like `takeN_` to be as fast as possible and for that reason I used `splitAt`. I was much more preoccupied by raw speed than by the potential danger of sharing long input streams. I'd be curious to know how adding `copy` affects the benchmarks. If it is not too bad, which it could well be, we might do better by just adding `copy` in the right places and avoid sharing.

Would you have time to add `copy` and run the benchmarks that are included with the library (before/after the change)?

EggBaconAndSpam:
Thanks for having a look at this!

I'll get you a PR / more data by the end of this week (this is mainly a deadline for myself to not postpone this indefinitely).

PR #495, opened 2022-10-28
Explicitly copy token sequences to avoid sharing input

EggBaconAndSpam:
Closes https://github.com/mrkkrp/megaparsec/issues/492

As mentioned in the issue, we can avoid leaking memory retaining the input in memory by explicitly copying token sequences as we read them.

The performance implications of adding an explicit `copy` to `takeN_` and `takeWhile_` are as follows:
- The `takeWhileP` family of basic combinators end up consuming a lot more memory (since previously they didn't need to allocate the result buffer) and time (due to the `copy` call).
- The `decimal` (`octal` etc) family of combinators are about 40% slower and consume about twice as much memory. 
- All other combinators (like `many`) have either stayed the same or got slightly (up to 10%) faster.

Sadly we don't have any real-world benchmarks (or do we?) so it's hard to tell if this would have any significant impact in practice.

### Raw benchmark results
This PR: [memory](https://github.com/mrkkrp/megaparsec/files/9885725/copy_memory.txt) [speed](https://github.com/mrkkrp/megaparsec/files/9885726/copy_speed.txt)
Master: [memory](https://github.com/mrkkrp/megaparsec/files/9885727/master_memory.txt) [speed](https://github.com/mrkkrp/megaparsec/files/9885728/master_speed.txt)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
   class allows us to use `dbg` in MTL monad transformers. [Issue
   488](https://github.com/mrkkrp/megaparsec/issues/488).
 
+* Introduced the `ShareInput` and `NoShareInput` newtype wrappers in
+  `Text.Megaparsec.Stream` in order to allow the user to choose how the
+  input should be sliced and shared during the parsing. [Issue
+  492](https://github.com/mrkkrp/megaparsec/issues/492).
+
 ## Megaparsec 9.2.2
 
 * Fixed a space leak in the implementations of the `reachOffset` and
--- a/Text/Megaparsec/Stream.hs
+++ b/Text/Megaparsec/Stream.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE Safe #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 -- |
 -- Module      :  Text.Megaparsec.Stream
@@ -25,11 +26,14 @@
 -- @since 6.0.0
 module Text.Megaparsec.Stream
   ( Stream (..),
+    ShareInput (..),
+    NoShareInput (..),
     VisualStream (..),
     TraversableStream (..),
   )
 where
 
+import Data.Bifunctor (second)
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.ByteString.Lazy as BL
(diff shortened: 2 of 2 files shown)


mrkkrp:
The way forward here is to first bring up to date the nix stuff that allows us to observe effects of such changes on dependent packages. I will do it soon. Once that is done we will be able to make an informed decision.

mrkkrp:
Here are the results of benchmark before (old) and after (new) the changes for 3 packages (`parsers-bench` from this repo, `mmark`, and `modern-uri`):

* [new-mmark-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043681/new-mmark-bench-memory.txt)
* [new-mmark-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043682/new-mmark-bench-speed.txt)
* [new-modern-uri-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043684/new-modern-uri-bench-memory.txt)
* [new-modern-uri-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043685/new-modern-uri-bench-speed.txt)
* [new-parsers-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043687/new-parsers-bench-memory.txt)
* [new-parsers-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043688/new-parsers-bench-speed.txt)
* [old-mmark-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043689/old-mmark-bench-memory.txt)
* [old-mmark-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043690/old-mmark-bench-speed.txt)
* [old-modern-uri-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043691/old-modern-uri-bench-memory.txt)
* [old-modern-uri-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043692/old-modern-uri-bench-speed.txt)
* [old-parsers-bench-memory.txt](https://github.com/mrkkrp/megaparsec/files/10043693/old-parsers-bench-memory.txt)
* [old-parsers-bench-speed.txt](https://github.com/mrkkrp/megaparsec/files/10043694/old-parsers-bench-speed.txt)

While the impact sometimes is not big, in certain cases it is quite noticeable. Perhaps we should give the users a choice whether to perform copying and avoid sharing?

EggBaconAndSpam:
Thanks for sorting out the benchmarks! I had completely overlooked that `parsers-bench` is part of this repo as well :sweat_smile: 

I agree the impact appears to be large enough that we probably should give people a choice.

I propose adding two newtype wrappers `ShareInput` and `CopyInput` (names are up for debate :smile:) to explicitly select one or the other `Stream` implementation. The unwrapped instances should coincide with their `ShareInput` counterparts at least for now, i.e. `instance Stream a` = `instance Stream (ShareInput a)` in handwavy terms.

In a future major release we could change the default behaviour to `CopyInput`, with a short note on what input sharing is and to use the `ShareInput` instance if performance is a concern.

What do you think? Shall I update the PR along those lines?

mrkkrp:
This sounds good, please go ahead.

EggBaconAndSpam:
I added `ShareInput` and `NoShareInput`. What does this look like to you? :slightly_smiling_face:

mrkkrp (approved):

mrkkrp:
Thanks!

PR #501, opened 2022-12-08
add HLS to shell env

yaitskov:
I started looking into megaparsec development and noticed that IDE (`code .`) inside `nix develop` doesn't have a language server.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -141,7 +141,8 @@
         megaparsec = sdistTarball haskellPackages.megaparsec;
         megaparsec-tests = sdistTarball haskellPackages.megaparsec-tests;
       };
-
+      haskellLanguageServer = pkgs.haskell.lib.overrideCabal haskellPackages.haskell-language-server
+        (_: { enableSharedExecutables = true; });
     in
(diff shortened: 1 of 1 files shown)


mrkkrp:
I'm a bit ambivalent about this one. I usually have HLS globally installed.

flake.nix
```diff
@@ -141,7 +141,8 @@
         megaparsec = sdistTarball haskellPackages.megaparsec;
         megaparsec-tests = sdistTarball haskellPackages.megaparsec-tests;
       };
-
+      haskellLanguageServer = pkgs.haskell.lib.overrideCabal haskellPackages.haskell-language-server
```
mrkkrp, line 144: Why is this necessary?

yaitskov: Nixshell provides HLS, so `code .` in shell gets proper HLS for GHC.

mrkkrp: My question was rather specifically why `enableSharedExecutables = true` should be set, why not use `haskellLanguageServer` as it is?

mrkkrp:
Closing due to lack of activity (it doesn't seem to be very important and my last question is still not answered).

PR #514, opened 2023-02-28
Implement mkParsec—a new method of MonadParsec

### LINKED ISSUE #366: Add a new primitive: 'breakOn'
https://github.com/mrkkrp/megaparsec/blob/87d425759ab5c290c1cb54b1d52c7bffaae1e1ff/Text/Megaparsec/Stream.hs#L291

Would you consider adding a primitive combinator for `breakOn`? Or is that something you've already discussed before?

http://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-Lazy.html#v:breakOn

Something like `takeUpto`?

If so, I might be able to open a pull request.

brandonchinn178:
Close #366.

Implements a new primitive as discussed in https://github.com/mrkkrp/megaparsec/issues/366#issuecomment-1434839814, which enables defining new primitives, at the cost of needing to drop (a little) into megaparsec internals.

### DIFF
--- a/Text/Megaparsec/Class.hs
+++ b/Text/Megaparsec/Class.hs
@@ -37,6 +37,7 @@ import qualified Control.Monad.Trans.Writer.Lazy as L
 import qualified Control.Monad.Trans.Writer.Strict as S
 import Data.Set (Set)
 import Text.Megaparsec.Error
+import {-# SOURCE #-} Text.Megaparsec.Internal (Reply)
 import Text.Megaparsec.State
 import Text.Megaparsec.Stream
 
@@ -272,6 +273,13 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
   -- | @'updateParserState' f@ applies the function @f@ to the parser state.
   updateParserState :: (State s e -> State s e) -> m ()
 
+  -- | An escape hatch for defining custom 'MonadParsec' primitives. You
+  -- will need to import "Text.Megaparsec.Internal" in order to construct
+  -- 'Reply'.
+  --
+  -- @since 9.4.0
+  mkParsec :: (State s e -> Reply e s a) -> m a
+
 ----------------------------------------------------------------------------
 -- Lifting through MTL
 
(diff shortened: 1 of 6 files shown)


Text/Megaparsec/Class.hs
```diff
@@ -272,6 +273,29 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
...
+  -- @
+  -- breakOn :: Text -> ParsecT e Text m a
+  -- breakOn end = liftConsumer $ \\s ->
+  --   let (pre, post) = T.breakOn end (stateInput s)
+  --    in ( ConsumedOk mempty pre
+  --       , s
+  --           { stateInput = post
+  --           , stateOffset = stateOffset s + T.length pre
+  --           }
+  --       )
```
brandonchinn178, line 288: Is this a valid implementation? When would `statePosState` need to be updated?

Text/Megaparsec/Class.hs
```diff
@@ -272,6 +273,29 @@ class (Stream s, MonadPlus m) => MonadParsec e s m | m -> e s where
...
+  -- @
+  liftConsumer :: (State s e -> (ParseResult s e a, State s e)) -> m a
+
+-- | A helper containing all the information that 'liftConsumer' can return.
+data ParseResult s e a
+  = ConsumedOk (Set (ErrorItem (Token s))) a
+  | ConsumedErr (ParseError s e)
+  | EmptyOk (Set (ErrorItem (Token s))) a
+  | EmptyErr (ParseError s e)
```
brandonchinn178, line 297: This could also be
```hs
(Either (ParseError s e) (Set (ErrorItem (Token s))), a), Bool)
```
But I figured a little duplication + 1 more data type is much nicer than that

Lev135 (commented):
Why not to use `Reply`? It seems that your type of liftConsumer is isomorphic to
```
(State s e -> Reply s e a) -> m a
```
and implementation of `pLiftConsumer` is identical to `mkPT`.

mrkkrp:
@brandonchinn178 So, first of all, I'm sorry it took so long to get to this one. This is great idea which I believe can greatly empower users of the library. Regarding @Lev135's suggestion, I think it is a very good one. I sketched an alternative implementation and pushed it to this branch (I also took the liberty of rebasing on current `master`). Can you please check that this one `mkParsec` method is sufficient for doing whatever you did with `liftConsumer`?

I did not implement `breakOn` just yet, but I you deem `mkParsec` usable we can add the example with `breakOn` back and merge this.

brandonchinn178:
Looks great! Yeah, if you get the breakOn test working, that's sufficient for me!

mrkkrp (approved):

PR #530, opened 2023-06-05
Print hints in `dbg` method

Lev135:
This is convinient for figuring out where your labels are lost during parsing process.
I had to add `showErrorItem` to `Text.Megaparsec.Error` exports to use it for printing hints in `Text.Megaparsec.Debug`. Maybe there's any better solution. Moreover, there can be many styles of printing hints. I've chosen list-like:
```
HINTS: ['a', 'b', end of file]
```

### DIFF
--- a/Text/Megaparsec/Error.hs
+++ b/Text/Megaparsec/Error.hs
@@ -43,6 +43,7 @@ module Text.Megaparsec.Error
     errorBundlePretty,
     parseErrorPretty,
     parseErrorTextPretty,
+    showErrorItem,
   )
 where
 
@@ -458,6 +459,8 @@ parseErrorTextPretty (FancyError _ xs) =
 -- Helpers
 
 -- | Pretty-print an 'ErrorItem'.
+--
+-- @since 9.4.0
 showErrorItem :: (VisualStream s) => Proxy s -> ErrorItem (Token s) -> String
 showErrorItem pxy = \case
   Tokens ts -> showTokens pxy ts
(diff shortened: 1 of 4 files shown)


Text/Megaparsec/Error.hs
```diff
@@ -43,6 +43,7 @@ module Text.Megaparsec.Error
     errorBundlePretty,
     parseErrorPretty,
     parseErrorTextPretty,
+    showErrorItem,
```
mrkkrp, line 46: You should add `@since 9.4.0` to the documentation of this function (let's say the next release will be 9.4.0).

megaparsec-tests/tests/Text/Megaparsec/DebugSpec.hs
```diff
@@ -179,6 +186,7 @@ spec = do
 -- | Check that running the given parser on the input prints the expected
 -- string to the 'stderr'.
 shouldStderr ::
+  (HasCallStack) =>
```
mrkkrp, line 189: Why would this become necessary now?

mrkkrp (commented):
Could you please also update the changelog accordingly?

megaparsec-tests/tests/Text/Megaparsec/DebugSpec.hs
Lev135: This is useful for searching failed tests in spec. Without this constraint error message refer to this function instead of appropriate test case

mrkkrp (approved):

mrkkrp:
Thanks!

PR #534, opened 2023-06-21
Add MonadWriter instance

olafklinke:
A general function `hoistP` can be used to implement monad transformer class methods of the form `MonadFoo m => m (f a) -> m (g a)` for example `local`, `listen` and `pass`. Usage of `hoistP` becomes more ergonomic when `Result` and `Reply` have `Traversable` instances.

### DIFF
--- a/Text/Megaparsec/Internal.hs
+++ b/Text/Megaparsec/Internal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
@@ -5,6 +6,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE Safe #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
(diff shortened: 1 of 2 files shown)


Text/Megaparsec/Internal.hs
```diff
@@ -100,6 +103,7 @@ instance (Ord t) => Monoid (Hints t) where
 --
 -- See also: 'Consumption', 'Result'.
 data Reply e s a = Reply (State s e) Consumption (Result s e a)
+  deriving (Functor)
```
mrkkrp, line 106: `Functor` instances can be automatically derived.

Text/Megaparsec/Internal.hs
```diff
@@ -249,12 +254,30 @@ instance (Stream s, MonadIO m) => MonadIO (ParsecT e s m) where
...
+  local f = hoistP (local f)
 
 instance (Stream s, MonadState st m) => MonadState st (ParsecT e s m) where
   get = lift get
   put = lift . put
 
+hoistP ::
+  (Stream s, Monad m) =>
+  (m (Reply e s a) -> m (Reply e s b)) ->
+  ParsecT e s m a ->
```
mrkkrp, line 266: Here, I opted for a less polymorphic version of the signature, since the extra flexibility is not actually needed, so it could confuse future readers (e.g. "do we actually perform a natural tranformation in these cases?").

Text/Megaparsec/Internal.hs
```diff
@@ -249,12 +254,30 @@ instance (Stream s, MonadIO m) => MonadIO (ParsecT e s m) where
...
+  ParsecT e s m b
+hoistP h p = mkParsecT (h . runParsecT p)
+
+-- | @since 9.5.0
+instance (Stream s, MonadWriter w m) => MonadWriter w (ParsecT e s m) where
+  tell w = lift (tell w)
+  listen = hoistP (fmap (\(repl, w) -> fmap (,w) repl) . listen)
+  pass = hoistP $ \m -> pass $ do
```
mrkkrp, line 274: While I do appreciate the clever point-free implementation using traversals and `Endo`, I think it is not the most readable one. I went ahead and rewrote it in a more verbose way.

mrkkrp (approved):

PR #539, opened 2023-07-16
Export initialState and initialPosState

### LINKED ISSUE #449: Export initialState
One useful feature that `megaparsec` has compared to `parsec` is the ability to get the remainder of the input. In fact, I think it is so useful that it would make sense to expose this functionality explicitly.

In the meantime, it is accessible via `runParser'`, however it is not easy to use given that this function requires the initial state, which is pretty tricky and boilerplate-y to construct. I think, it would be helpful to export `initialState` used by `runParser` in order to:

1. Better support the use case of getting the remaining input.
2. In general, simplify the initial state creation to allow the user to just override the fields in the default state that they want.

Lev135:
Resolves #449. I'm not sure if we should export some function constructing initial `PosState`. Names are also up to debate. These at least don't clash with `hspec-megaparsec`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 *Megaparsec follows [SemVer](https://semver.org/).*
 
+## Unreleased
+
+* Added the functions `initialState` and `initialPosState` to
+  `Text.Megaparsec.State`. [Issue
+  449](https://github.com/mrkkrp/megaparsec/issues/449).
+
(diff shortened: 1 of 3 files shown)


mrkkrp:
In the end I guess it is not that catastrophic to use the same names as in `hspec-megaparsec`, given that `Text.Megaparsec.State` is not re-exported from `Text.Megaparsec` and is rarely imported explicitly.

PR #547, opened 2023-10-13
Update for GHC 9.8 

amesgen:
Otherwise, one might get errors like
```
Error: cabal: Could not resolve dependencies:
[__0] trying: ormolu-0.7.2.0 (user goal)
[__1] trying: th-env-0.1.1 (dependency of ormolu)
[__2] trying: template-haskell-2.21.0.0/installed-54dc (dependency of th-env)
[__3] trying: pretty-1.1.3.6/installed-8ddc (dependency of template-haskell)
[__4] trying: megaparsec-9.5.0 (dependency of ormolu)
[__5] next goal: Cabal-syntax (dependency of ormolu)
[__5] rejecting: Cabal-syntax-3.10.2.0/installed-6d41 (conflict: megaparsec =>
bytestring>=0.2 && <0.12, Cabal-syntax => bytestring==0.12.0.2/installed-66e0)
[__5] rejecting: Cabal-syntax-3.10.1.0 (conflict: pretty =>
deepseq==1.5.0.0/installed-2833, Cabal-syntax => deepseq>=1.3.0.1 && <1.5)
[__5] rejecting: Cabal-syntax-3.8.1.0 (conflict: ormolu => Cabal-syntax>=3.10
&& <3.11)
[__5] skipping: Cabal-syntax-3.6.0.0 (has the same characteristics that caused
the previous version to fail: excluded by constraint '>=3.10 && <3.11' from
'ormolu')
[__5] fail (backjumping, conflict set: Cabal-syntax, megaparsec, ormolu,
pretty)
```
as `bytestring-0.12` (shipped with GHC 9.8) is non-reinstallable when using TH.

 - Allowing `containers-0.7` is not strictly necessary, but it compiles fine.
 - Due to this change in `base-4.19`
   
   > Added `{-# WARNING in "x-partial" #-}` to `Data.List.{head,tail}`.
   
   there are various new warnings with GHC 9.8. We fix these by enabling `-Wno-x-partial`.

### DIFF
--- a/megaparsec-tests/megaparsec-tests.cabal
+++ b/megaparsec-tests/megaparsec-tests.cabal
@@ -5,7 +5,7 @@ license:         BSD-2-Clause
 license-file:    LICENSE.md
 maintainer:      Mark Karpov <markkarpov92@gmail.com>
 author:          Megaparsec contributors
-tested-with:     ghc ==9.2.7 ghc ==9.4.4 ghc ==9.6.1
+tested-with:     ghc ==9.2.8 ghc ==9.4.7 ghc ==9.6.3 ghc ==9.8.1
 homepage:        https://github.com/mrkkrp/megaparsec
 bug-reports:     https://github.com/mrkkrp/megaparsec/issues
 synopsis:        Test utilities and the test suite of Megaparsec
--- a/megaparsec-tests/megaparsec-tests.cabal
+++ b/megaparsec-tests/megaparsec-tests.cabal
@@ -26,10 +26,10 @@ library
     build-depends:
         QuickCheck >=2.10 && <2.15,
         base >=4.15 && <5,
-        bytestring >=0.2 && <0.12,
-        containers >=0.5 && <0.7,
-        hspec >=2.0 && <3,
-        hspec-megaparsec >=2.0 && <3,
+        bytestring >=0.2 && <0.13,
+        containers >=0.5 && <0.8,
+        hspec >=2.0 && <3.0,
+        hspec-megaparsec >=2.0 && <3.0,
         megaparsec ==9.5.0,
         mtl >=2.2.2 && <3,
         text >=0.2 && <2.2,
@@ -64,11 +64,11 @@ test-suite tests
     build-depends:
         QuickCheck >=2.10 && <2.15,
         base >=4.15 && <5,
-        bytestring >=0.2 && <0.12,
+        bytestring >=0.2 && <0.13,
         case-insensitive >=1.2 && <1.3,
-        containers >=0.5 && <0.7,
+        containers >=0.5 && <0.8,
         hspec >=2.0 && <3,
-        hspec-megaparsec >=2.0 && <3,
+        hspec-megaparsec >=2.0 && <3.0,
         megaparsec ==9.5.0,
         megaparsec-tests,
         mtl >=2.2.2 && <3,
@@ -84,3 +84,6 @@ test-suite tests
 
     else
         ghc-options: -O2 -Wall
+
+    if impl(ghc >=9.8)
+        ghc-options: -Wno-x-partial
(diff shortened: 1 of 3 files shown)


Text/Megaparsec/Char/Lexer.hs
```diff
@@ -346,7 +346,7 @@ charLiteral = label "literal character" $ do
   r <- lookAhead (count' 1 10 anySingle)
   case listToMaybe (Char.readLitChar r) of
     Just (c, r') -> c <$ skipCount (length r - length r') anySingle
-    Nothing -> unexpected (Tokens (head r :| []))
+    Nothing -> unexpected (Tokens (NE.fromList (take 1 r)))
```
andreasabel, line 349: Isn't a use of `NE.fromList` as bad as using `head`?  Both will throw the exception in the _library_, not in the _application_ code.
If all that _head is bad_ excitement should lead to improvement, then the new code should be `fromMaybe undefined $ NE.nonEmpty $ take 1 r` which will raise the error (via `undefined`) here, not in the library.

andreasabel (approved):
Thanks. LGTM, with a question about `NE.fromList`.

andreasabel:
I published a revision of the bytestring bound to hackage: https://hackage.haskell.org/package/megaparsec-9.5.0/revisions/
This was missing for full GHC 9.8.1 compliancy.

Text/Megaparsec/Char/Lexer.hs
amesgen: Both `head` and `NE.fromList` use `HasCallStack`, so they will at least point to the code here. But yeah, I don't think there is any big advantage of using `NE.fromList` rather than `head` other than to avoid the warning (in case that is even desired).

megaparsec-tests/megaparsec-tests.cabal
```diff
@@ -84,3 +84,6 @@ test-suite tests
 
     else
         ghc-options: -O2 -Wall
+
+    if impl(ghc >=9.8)
```
mrkkrp, line 88: I'd rather do without this. Even though partial functions are "bad", sometimes they are the simplest and the most reasonable solution available.

megaparsec.cabal
```diff
@@ -68,23 +68,22 @@ library
         transformers >=0.4 && <0.7
 
     if flag(dev)
-        ghc-options: -Wall -Werror -Wpartial-fields -Wunused-packages
+        ghc-options:
+            -Wall -Werror -Wredundant-constraints -Wpartial-fields
```
mrkkrp, line 72: I vaguely remember that there used to be a good reason not to use `-Wredundant-constraints`, but if CI passes we can totally bring it back.

amesgen: It had false positives in 9.6.2, but this was fixed in 9.6.3.

megaparsec-tests/megaparsec-tests.cabal
amesgen: I am not sure I understand; as far as I am aware, there are two options when using `-Werror` (as is done on CI here):

 1. Adapt usage sites of `head`/`tail` to sth else (refactorings, or other partial functions). (In this PR, this is done in one place using `NE.fromList`.)
 2. Use `-Wno-x-partial` to silence the warnings. (This is done in this PR right here.)

Are you advocating for one of these, or sth else?

mrkkrp: Ah, I simply misread `-Wno-x-partial` as `-Wx-partial`, missed the `no`.

amesgen: Right, would you then also prefer using `-Wno-x-partial` in order to be able to revert the `NE.fromList` change back to `head`?

mrkkrp: Yes, that would be nice.

mrkkrp:
Thanks!

PR #548, opened 2023-10-14
Formal correction of dep upper bounds: `< x.0` to `< x`

andreasabel:
The versions `5.0` and `5` are not the same.
If you write `base < 5.0`, you accept `base-5` but reject `base-5.0.*`.
This is usually not intended.
The correct format for upper bounds is `< ver` where `ver` has no trailing 0s.


mrkkrp:
Hi, thanks! Could you apply the same fix for `megaparsec-tests`?

andreasabel:
> Hi, thanks! Could you apply the same fix for `megaparsec-tests`?

Done.

mrkkrp:
Thanks!

PR #549, opened 2023-10-29
Move `Text.Megaparsec.State` into `exposed-modules`

byorgey:
#539 exported `initialState` and `initialPosState`, and the CHANGELOG for 9.6.0 indicated that these are now available for use.  Unfortunately, they are not, since they were exported from `Text.Megaparsec.State`, which itself is not exported from the package.  This PR simply moves `Text.Megaparsec.State` from `other-modules` to `exported-modules` to allow `initialState` and `initialPosState` to be used as advertised.

If another solution is considered more desirable (*e.g.* re-exporting `initialState` and `initialPosState` from some other module) I'd be happy to modify the PR or open a new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 *Megaparsec follows [SemVer](https://semver.org/).*
 
+## Megaparsec 9.6.1
+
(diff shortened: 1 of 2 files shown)


mrkkrp:
Thanks!

PR #559, opened 2024-05-16
Expand elem/notElem to allow fusion

1Jajen1:
GHC doesn't seem to inline or fuse the infix version of `elem/notElem`.  Fully expanding their uses does allow it.

csv benchmarks get faster and `satisfy (\x -> x == '\'' || x == '\"') = oneOf ['\'', '\"']` now holds with regards to performance. Note that `oneOf "'\""` will still be slower, which is why I left the note unchanged.

### DIFF
--- a/Text/Megaparsec.hs
+++ b/Text/Megaparsec.hs
@@ -516,7 +516,7 @@ oneOf ::
   -- | Collection of matching tokens
   f (Token s) ->
   m (Token s)
-oneOf cs = satisfy (`elem` cs)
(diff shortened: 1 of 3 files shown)


mrkkrp (approved):

mrkkrp:
Thanks! Strangely, I see this

> This branch cannot be rebased due to too many changes

and the merge button is grey. As far as I can see this is based directly on current master, so it is not clear how there could be "too many changes".

PR #566, opened 2024-08-06
Inline applicative operators

szumixie:
The applicative operators `*>` and `<*` are not inlined in the generated GHC Core, so I added `INLINE` pragmas to them. As a result, parsing CSV and JSON takes around 40% less time when running the benchmarks on my machine.


mrkkrp (approved):

mrkkrp:
Thanks!

PR #603, opened 2026-06-18
Fix handling of zero-width characters in error messages

### LINKED ISSUE #572: Error is formatted incorrectly when 0-width characters are in use.
I know this is nitpicky but I felt it important to point out in case this is supposed to be supported.

Here is an example error:

```
Failed to parse message:
1:1:
  |
1 | <empty line>
  | ^
<pure>:1:64:
  |
1 | 34=149=XXXX_XXX_XXXX52=20240930-15:03:41.83656=XXXXX.TEST57=XXXXX.XXX298=0108=3010=198
  |                                                                ^
Unknown field tag: 57
```

As you can see, the `^` is pointing at the wrong character.
I think this is because there are invisible (0-width) `\SOH` characters inbetween fields.

mrkkrp:
Close #572.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 *Megaparsec follows [SemVer](https://semver.org/).*
 
-## Megaparsec 9.7.2
+## Megaparsec 9.8.0
 
 * Fixed the associativity of the `(<|>)` operator. [Issue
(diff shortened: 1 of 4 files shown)


NorfairKing:
Thank you :hugs:

PR #604, opened 2026-06-18
Require base-4.20

Bodigrim:
Earlier versions of `base` can no longer be used to build `megaparsec-9.8.0` because of the lack of `Prelude.foldl'`.

See https://hackage.haskell.org/package/megaparsec-9.8.0/reports/2 for an example of build failure with `base-4.19`. 

Could you please make a revision for `megaparsec-9.8.0`?


mrkkrp:
Done, I made revisions for both `megaparsec-9.8.0` and `megaparsec-tests-9.8.0`.

PR #606, opened 2026-06-23
Fix the regression with the pointer marker ^ not being shown anymore

mrkkrp:
@simonmichael Can you please take a look and confirm that this solves the issue for you?


simonmichael:
Confirmed! Thank you @mrkkrp.

